PR #1, opened 2020-09-23
Add initial docker-compose.yml

jakobw:
This adds an initial docker-compose.yml for running npm commands. This is copy-pasta of the relevant bits from one of our existing code bases using a similar tech stack, e.g. our design system.

Wikit's docker-compose.yml: https://github.com/wmde/wikit/blob/master/docker-compose.yml

Phabricator: [T263542](https://phabricator.wikimedia.org/T263542)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,16 @@
+FROM docker-registry.wikimedia.org/nodejs10-devel
+
+RUN apt-get update && \
+	apt-get install -y \
+		ca-certificates \
+		build-essential \
+		python-pkgconfig \
+		git
+
+ARG UID=1000
+ARG GID=1000
+
+RUN addgroup --gid $GID runuser && adduser --uid $UID --gid $GID --disabled-password --gecos "" runuser
+RUN npm install --global npm@6.14.8
+
+USER runuser
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,11 @@
+version: '2'
+
+services:
+    node:
+        build:
+            context: ./
+        image: wmde/query-builder
+        volumes:
+            - '~/.npm:/.npm'
+            - './:/app'
+        working_dir: /app
--- a/README.md
+++ b/README.md
@@ -1,2 +1,19 @@
 # query-builder
 🔍️ A simple query builder for Wikidata SPARQL queries
+
+## Development
+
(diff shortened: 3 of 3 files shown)


Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+		ca-certificates \
+		build-essential \
+		python-pkgconfig \
+		git
+
+ARG UID=1000
+ARG GID=1000
+
+RUN addgroup --gid $GID runuser && adduser --uid $UID --gid $GID --disabled-password --gecos "" runuser
+RUN npm install --global npm@6.14.5
```
wiese:
```suggestion
RUN npm install --global npm@6.14.8
```

(if we don't do it now it will stay like this for a while)

Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+
```
wiese, line 15: ~~Lerna may not be needed~~ (diff changed)

Dockerfile
jakobw: Good point, thanks. Done!

Dockerfile
jakobw: Yeah, I was a bit too generous with the copying :spaghetti:

docker-compose.yml
```diff
@@ -0,0 +1,12 @@
+version: '2'
+
+services:
+    node:
+        build:
+            context: ./
+            dockerfile: ./Dockerfile
```
bereket-WMDE: If you don't have a custom name docker file, do we need to specify it in the "dockerfile" section ? 
It think it should automatically pick it up based on the 'context'

wiese (commented):
LGTM but not approving as I don't quite know (not part of the team) where this is going.

I assume in a next step some form of service extending `node` will be added - which would mitigate the often heard criticism that having docker-compose is overkill if there are no actual services (beyond a cli run) and nothing is composed.

docker-compose.yml
wiese: Probably an artifact from https://gerrit.wikimedia.org/r/plugins/gitiles/wikibase/termbox/+/89dd6e5/docker-compose.yml#7 (a possible explanation but not a good justification for the status quo here and in wikit)

jakobw: It is indeed not needed. Removed it!

bereket-WMDE (approved):

Ladsgroup (approved):

PR #3, opened 2020-09-23
Add docker-compose dev service

jakobw:
This adds a service for running the development server for the
application with a configurable port via the `PORT` environment
variable.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,0 +1,2 @@
+# The port for the development server
+PORT=8080
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@ node_modules
 
 
 # local env files
+.env
 .env.local
 .env.*.local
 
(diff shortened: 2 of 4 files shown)


.env
```diff
@@ -0,0 +1,2 @@
+# The port for the development server
+PORT=8080
```
bereket-WMDE: should be added to .gitignore file

jakobw: Strange. I thought I've seen a working example of a checked in `.env` file for default environment variable values that can be overridden by `.env.local` and the likes. This does indeed not seem to work as I imagined, though. Changing it, thanks!

jakobw: Changed it and added an `.env.example` instead.

wiese: The more you know. I also assumed this would do (and acted upon this assumption in the past) being used to e.g.
https://github.com/bkeepers/dotenv#what-other-env-files-can-i-use
https://symfony.com/doc/current/configuration.html#overriding-environment-values-via-env-local

jakobw: Yeah, interestingly the webpack dev server happily picked up the `PORT` variable from the `.env.local`, but that obviously didn't help with mapping it to the host's port.

bereket-WMDE (approved):
tested it. works !

bereket-WMDE (approved):
works

PR #5, opened 2020-09-23
Enable Wikimedia eslint configs

jakobw:
This enables the [Wikimedia eslint config](https://github.com/wikimedia/eslint-config-wikimedia) (in a slightly outdated version due to dependency conflicts) as well as [WMDE's typescript eslint config](https://github.com/wmde/eslint-config-wikimedia-typescript#readme). Unfortunately I found many of Prettier's rules to conflict with the aforementioned configs and removed it for now. We can try again installing Prettier at a later point in time, but redundantly specifying the rules in the eslint config and the prettier config seems silly to me.


Ladsgroup (approved):
LGTM, my only comment is that we are cleaning the Vue boilerplate while at end of the day we are going to throw them away but it's nitpick and I assume it's needed to make CI green on master.

jakobw:
Latest push is a conflict-less rebase, so I assume it's ok to merge!

PR #6, opened 2020-09-24
updated installation steps in right order

guergana:


guergana:
@jakobw @bereket-WMDE and I pair programmed on this. Will merge.

PR #10, opened 2020-09-25
Add ADR0000

Ladsgroup:
Use Toolforge kubernetes infrastructure for the test system

Bug: T263539

### DIFF
--- a/docs/adr/0000-use-toolforge-kubernetes-for-test-system.md
+++ b/docs/adr/0000-use-toolforge-kubernetes-for-test-system.md
@@ -21,7 +21,7 @@ We are going to have the test setup in Toolforge, served as static assets but bu
 
 ## Consequences
 
-This while taking advantage of modern technologies like kubernetes jobs (instead of Gridengine jobs) is fairly simple and straightforward, especially compared to having a dedicated VM and serving these as either a nodejs app or using Apache/Nginx. This can also be used in a similar manner for other simple client-side apps we will develop.
+This, while taking advantage of modern technologies like kubernetes jobs (instead of Gridengine jobs), is fairly simple and straightforward, especially compared to having a dedicated VM and serving these as either a nodejs app or using Apache/Nginx. This can also be used in a similar manner for other simple client-side apps we will develop.
 
 It means at most, it would take an hour for a patch to be actually visible in the test system and it can be modified to be faster or slower depending on the need.
 
--- a/docs/adr/0000-use-toolforge-kubernetes-for-test-system.md
+++ b/docs/adr/0000-use-toolforge-kubernetes-for-test-system.md
@@ -17,7 +17,7 @@ Toolforge seems like a good setup for small projects due to its baked-in support
 
 ## Decision
 
-We are going to have the test setup in Toolforge, served as static assets but built using Kubernetes cronjob that builds these assets using node10 docker container in Toolforge automatically and copies them to the public endpoint. You can see the exact configuration and steps in [PR#9](https://github.com/wmde/query-builder/pull/9).
+We are going to have the test setup in Toolforge, served as static assets but built using a Kubernetes cronjob that builds these assets using the node10 docker image in Toolforge automatically and copies them to the public endpoint. You can see the exact configuration and steps in [PR#9](https://github.com/wmde/query-builder/pull/9).
 
 ## Consequences
 
--- a/docs/adr/0000-use-toolforge-kubernetes-for-test-system.md
+++ b/docs/adr/0000-use-toolforge-kubernetes-for-test-system.md
@@ -4,7 +4,7 @@ Date: 2020-09-25
 
 ## Status
 
-proposed
+accepted
 
 ## Context
 
--- a/docs/adr/0000-use-toolforge-kubernetes-for-test-system.md
+++ b/docs/adr/0000-use-toolforge-kubernetes-for-test-system.md
@@ -10,7 +10,7 @@ accepted
 
 This project is new and its exact technological future is unknown. For example, we still don't know if this will have to be a nodejs app in the future, or if it will be only a client-side app served as static assets using Apache, Nginx, etc.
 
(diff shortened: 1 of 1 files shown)


docs/adr/0000-use-toolforge-kubernetes-for-test-system.md
```diff
@@ -0,0 +1,28 @@
+# 0) Use Toolforge kubernetes infrastructure for the test system
+
+Date: 2020-09-25
+
+## Status
+
+proposed
```
jakobw:
```suggestion
accepted
```
!

docs/adr/0000-use-toolforge-kubernetes-for-test-system.md
```diff
@@ -0,0 +1,28 @@
...
+
+## Context
+
+This project is new and its exact technological future is unknown, for example, we still don't know if this will have to be nodejs app in future and it will be only client-side app being built from npm modules and served as static assets using Apache, Nginx, etc.
```
jakobw:
```suggestion
This project is new and its exact technological future is unknown. For example, we still don't know if this will have to be a nodejs app in the future, or if it will be only a client-side app served as static assets using Apache, Nginx, etc.
```

docs/adr/0000-use-toolforge-kubernetes-for-test-system.md
```diff
@@ -0,0 +1,28 @@
...
+
+We also need a publicly accessible test setup to have its features tested by UX and PM once they land on the main branch. We also need a development environment, need a CI environment and so on. Each environment has its own unique requirements, for production we are not sure we would use pipeline as a node app or it will be part of the standard WDQS GUI assets.
```
jakobw:
```suggestion
We also need a publicly accessible test setup to have its features tested by UX and PM once they land on the main branch. We also need a development environment, need a CI environment and so on. Each environment has its own unique requirements. For production we are not sure yet whether we will use [the Blubber deployment pipeline](https://wikitech.wikimedia.org/wiki/Blubber) and deploy the Query Builder as a node app or whether it will be part of the standard WDQS GUI assets.
```

docs/adr/0000-use-toolforge-kubernetes-for-test-system.md
```diff
@@ -0,0 +1,28 @@
...
+
+Toolforge seems like a good setup for small projects due to its baked-in support of serving static assets using lighttpd or serving node10 apps. It has been changing and growing more mature. One of its most recent features is support for dockerized webservices and cronjobs using kubernetes. You can read more about it in [here](https://wikitech.wikimedia.org/wiki/Help:Toolforge/Kubernetes). We can't use our built Dockerfile in Toolforge and if we want to go in that direction, we would have to have a dedicated VM just for this service in CloudVPS instead and make it automatically updated using puppet or ansible which would be too much work for too little gain.
```
jakobw, line 15: > We can't use our built Dockerfile in Toolforge

I also had this in the back of my mind but reading this I was immediately wondering "why?!" because we are using a docker image from the WMF registry even for our development setup. Do you know if there's a piece of documentation that we can link to which explains this? If the reason is that only a fixed list of images is allowed  there, then it may be as simple as linking to that list.

docs/adr/0000-use-toolforge-kubernetes-for-test-system.md
```diff
@@ -0,0 +1,28 @@
...
+
+
+## Decision
+
+We are going to have the test setup in Toolforge, served as static assets but built using Kubernetes cronjob that builds these assets using node10 docker container in Toolforge automatically and copies them to the public endpoint. You can see the exact configuration and steps in [PR#9](https://github.com/wmde/query-builder/pull/9).
```
jakobw:
```suggestion
We are going to have the test setup in Toolforge, served as static assets but built using a Kubernetes cronjob that builds these assets using the node10 docker image in Toolforge automatically and copies them to the public endpoint. You can see the exact configuration and steps in [PR#9](https://github.com/wmde/query-builder/pull/9).
```

docs/adr/0000-use-toolforge-kubernetes-for-test-system.md
```diff
@@ -0,0 +1,28 @@
...
+
+## Consequences
+
+This while taking advantage of modern technologies like kubernetes jobs (instead of Gridengine jobs) is fairly simple and straightforward, especially compared to having a dedicated VM and serving these as either a nodejs app or using Apache/Nginx. This can also be used in a similar manner for other simple client-side apps we will develop.
```
jakobw:
```suggestion
This, while taking advantage of modern technologies like kubernetes jobs (instead of Gridengine jobs), is fairly simple and straightforward, especially compared to having a dedicated VM and serving these as either a nodejs app or using Apache/Nginx. This can also be used in a similar manner for other simple client-side apps we will develop.
```

jakobw (changes requested):
Thanks for writing this up! Looks good to me! I mostly nitpicked around in the text and added one question as to why we're limited in our choice of docker images on Toollforge which I think would be nice to have covered.

docs/adr/0000-use-toolforge-kubernetes-for-test-system.md
Ladsgroup: I can ask around (and I couldn't find such documentation) but I assume it's because any arbitrary Dockerfile gives a big degree of freedom which could become problematic (noisy neighbour effect).

Ladsgroup: The answer I got in IRC:
> Amir1: you would then need to upload the resulting docker img to our toolforge registry, right? We don't support bring-your-own-container yet, even if it is based on the WMF docker registry

jakobw (approved):
LGTM :sparkles: 

I'd still be curious to hear from @bereket-WMDE whether this ADR answers [all of his questions](https://phabricator.wikimedia.org/T263539#6491553).

bereket-WMDE:
> LGTM sparkles
> 
> I'd still be curious to hear from @bereket-WMDE whether this ADR answers [all of his questions](https://phabricator.wikimedia.org/T263539#6491553).

yes it does. I've some discussions about this with @Ladsgroup  and adam about toolforge and its benefits. I was under the impression that toolforge is some VM and that was while i asked about the pipelines

PR #11, opened 2020-09-29
Start the tile and basic text part of the app

Ladsgroup:
This doesn't have much in it, just building the header of the app
outlined in the mockup

Bug: T263564


src/components/QueryBuilder.vue
```diff
@@ -0,0 +1,58 @@
...
+</script>
+
+<style scoped lang="scss">
+@import '~@wmde/wikit-tokens/dist/variables';
+
+.querybuilder {
+	position: relative;
+	width: 1024px;
+	height: 349px;
+	padding: 40px;  // $dimension-spacing-xlarge
```
guergana, line 24: Should we use here the new spacing tokens???

Ladsgroup: It's not released yet AFAIK. I couldn't use it directly, that's why I commented it so we can replace it once it's usable.

src/components/QueryBuilder.vue
```diff
@@ -0,0 +1,58 @@
...
+} );
+</script>
+
+<style scoped lang="scss">
+@import '~@wmde/wikit-tokens/dist/variables';
+
+.querybuilder {
+	position: relative;
+	width: 1024px;
+	height: 349px;
```
jakobw:
```suggestion
```

Those aren't part of the specs are far as I can tell. If I'm not mistaken the box should just grow with the window size.

src/components/QueryBuilder.vue
```diff
@@ -0,0 +1,58 @@
...
+}
+
+.querybuilder__heading {
+	font-family: $font-family-style-heading-serif;
+	font-weight: $font-weight-style-h1;
+	font-size: $font-size-style-h1;
+	line-height: $font-line-height-style-heading;
+	display: flex;
+	align-items: center;
+
```
jakobw:
```suggestion
```

I don't think this has any effect

tests/unit/querybuilder.spec.ts
```diff
@@ -0,0 +1,9 @@
+import { shallowMount } from '@vue/test-utils';
```
jakobw: Nitpick: can we give the test file the same upper camel case as the component? So `QueryBuilder.spec.ts`.

tests/unit/querybuilder.spec.ts
```diff
@@ -0,0 +1,9 @@
...
+import QueryBuilder from '@/components/QueryBuilder.vue';
+
+describe( 'QueryBuilder.vue', () => {
+	it( 'renders props.msg when passed', () => {
+		const wrapper = shallowMount( QueryBuilder, {} );
+		expect( wrapper.text() ).toMatch( '' );
+	} );
```
jakobw: Not that this test is super useful either way, but the test message threw me off. How about something like:

```suggestion
	it( 'has a heading', () => {
		const wrapper = shallowMount( QueryBuilder, {} );
		expect( wrapper.find( 'h1' ).text() ).toBe( 'Simple Query Builder' );
	} );
```

src/components/QueryBuilder.vue
```diff
@@ -0,0 +1,58 @@
+<template>
+	<div class="querybuilder">
+		<h1 class="querybuilder__heading">Simple Query Builder</h1>
+		<p class="querybuilder__description">
+			Short explanatory text that also manages expectations. Short explanatory text that also
+			manages expectations.
+		</p>
+		<h4 class="querybuilder__find-title">Find all items...</h4>
```
jakobw: This should be an `<h2>` even if the h4 tokens are used, I think.

jakobw (changes requested):
Looks pretty good! It's great to see the tokens in use in the SCSS :)
Found some small things to nitpick about as usual.

tests/unit/querybuilder.spec.ts
Ladsgroup: Yeah, I honestly forgot to fix it *hides*

tests/unit/querybuilder.spec.ts
Ladsgroup: Yeah I should do it.

src/components/QueryBuilder.vue
Ladsgroup: Yeah, I copied it from figma but probably I don't know how to copy things properly

src/components/QueryBuilder.vue
Ladsgroup: In the figma there's nothing and only it's the css
image

I'm fine either way, maybe we should double check with the designers?

tests/unit/querybuilder.spec.ts
Ladsgroup: Done!

src/components/QueryBuilder.vue
jakobw: I think this should only apply to the styling. As far as semantic markup goes the recommendation is pretty clear:

> Avoid skipping heading levels: always start from `<h1>`, next use `<h2>` and so on.

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Heading_Elements

Ladsgroup: oh thanks then. I fix it.

jakobw (approved):
Looks good! Squash'n'merge! :rocket:

PR #12, opened 2020-10-01
Fixes to make Toolforge test system to update again

Ladsgroup:
- Increasing the memory, with the default 0.5Gi, the cron constantly
   OOMs
 - Removing package lock file as it seems that causes error (likely due
   to different npm versions)

With these changes the test system is updating it again.

### DIFF
--- a/build/updater.sh
+++ b/build/updater.sh
@@ -2,7 +2,7 @@
 cd query-builder
 git pull
 rm -rf node_modules
-npm install npm@6.14.8
+npm install -g npm@6.14.8
 npm install
 npm run build
 rm -rf public_html/*
--- a/build/updater.sh
+++ b/build/updater.sh
@@ -2,7 +2,7 @@
 cd query-builder
 git pull
 rm -rf node_modules
-rm package-lock.json
+npm install npm@6.14.8
 npm install
 npm run build
 rm -rf public_html/*
--- a/build/updater.sh
+++ b/build/updater.sh
@@ -2,6 +2,7 @@
 cd query-builder
 git pull
 rm -rf node_modules
+rm package-lock.json
 npm install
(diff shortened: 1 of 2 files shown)


build/updater.sh
```diff
@@ -2,6 +2,7 @@
 cd query-builder
 git pull
 rm -rf node_modules
+rm package-lock.json
```
jakobw: We shouldn't do this. We [intentionally check in the package-lock.json](https://stackoverflow.com/questions/44206782/do-i-commit-the-package-lock-json-file-created-by-npm-5) to ensure that our different environments (dev, test system, prod) install the same dependencies. Removing it defeats the purpose of checking it in. If the NPM version is a problem, can't we just install the right one within the container? This is [what our Dockerfile does](https://github.com/wmde/query-builder/blob/master/Dockerfile#L14), too.

jakobw (changes requested):

build/updater.sh
Ladsgroup: Oh, It errored when installing and google said to delete the file. It actually makes a lot of sense. Let me give it a try.

Ladsgroup: Tested and it worked just fine. Changed now. Thanks for the hint.

build/updater.sh
```diff
@@ -2,6 +2,7 @@
 cd query-builder
 git pull
 rm -rf node_modules
+npm install npm@6.14.8
```
jakobw: I think this should say
```suggestion
npm install -g npm@6.14.8
```
to have the desired effect.

jakobw (changes requested):

jakobw (approved):
:pray:

PR #14, opened 2020-10-02
Use correct npm version to produce package-lock.json

Ladsgroup:
It was a mistake done by yours truly.


package.json
```diff
@@ -15,11 +15,12 @@
   },
   "main": "index.js",
   "dependencies": {
+    "@wmde/wikit-tokens": "1.0.1",
+    "@wmde/wikit-vue-components": "1.0.1",
     "core-js": "^3.6.5",
+    "npm": "^6.14.8",
```
jakobw: This line shouldn't be there. The version is fixed in the Dockerfile and npm is not a run-time dependency either.

jakobw (changes requested):

package.json
Ladsgroup: Fixed now.

jakobw:
Went in with #16

PR #17, opened 2020-10-05
Make tokens available for all components

jakobw:
This loads the tokens via sass-loader which means that we no longer have
to import them for every component that we want to use. It also allows
us to import components from @wmde/wikit-vue-components from the source
files without additional work to load the styles.


guergana (approved):
Tested with the button component. Looks good!

PR #18, opened 2020-10-05
Set up store and wire it to text input component

bereket-WMDE:
Bug: T264318

### DIFF
--- a/src/components/QueryBuilder.vue
+++ b/src/components/QueryBuilder.vue
@@ -10,10 +10,14 @@
 			<TextInput
 				class="querybuilder__rule__property"
 				label="Property"
+				:value="propertyLabel"
 				placeholder="Enter a property" />
 			<TextInput
 				class="querybuilder__rule__value"
 				label="Value"
+				ref="value"
+				:value="textInputValue"
+				@input="updateInputTextValue"
 				placeholder="Enter a value" />
 		</div>
 		<div class="querybuilder__run">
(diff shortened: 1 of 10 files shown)


src/components/QueryBuilder.vue
```diff
@@ -22,9 +25,25 @@
...
 import TextInput from '@wmde/wikit-vue-components/src/components/TextInput.vue';
+import { mapState } from 'vuex';
 
 export default Vue.extend( {
 	name: 'QueryBuilder',
+
+	data() {
+		return {
+			textInputValue: '',
+			textInputProperty: ''
```
jakobw: Neither of these are actually used. I think the `textInputProperty` one can just be deleted, and the value TextInput should actually read from the vuex store in the same way that the property one does, i.e. mapped it to a computed property from the store state.

jakobw (changes requested):
Looks good! I only found one small thing that we should take care of before merging.

jakobw (approved):
Needs some commit history cleanup, but otherwise looks great! Woo! :rocket:

guergana:
Looks good to me too. Just resolve the conflicts. Looks like it needs a rebase

jakobw (approved):
:100:

bereket-WMDE:
> Looks good! I only found one small thing that we should take care of before merging.

already fixed the merging issue

PR #19, opened 2020-10-05
Sync the updater.sh with the one in ToolForge

Ladsgroup:
- Installing npm locally and use its binary instead
 - Instead of dropping everything by deleting node_modules which would
   make the build really long (sometimes an hour or two), update it
   instead.
 - Do git reset --hard ot make sure the git repo is clean

### DIFF
--- a/build/updater.sh
+++ b/build/updater.sh
@@ -2,8 +2,8 @@
 cd /data/project/query-builder-test/query-builder
 git reset --hard origin/master
 git pull
-./node_modules/.bin/npm install npm@6.14.8
-./node_modules/.bin/npm update
+npm install npm@6.14.8
+./node_modules/.bin/npm install
 ./node_modules/.bin/npm run build
 rm -rf public_html/*
 cp -R dist/* ../public_html/
--- a/build/updater.sh
+++ b/build/updater.sh
@@ -1,9 +1,9 @@
 #!/bin/bash
-cd query-builder
+cd /data/project/query-builder-test/query-builder
+git reset --hard origin/master
 git pull
-rm -rf node_modules
-npm install -g npm@6.14.8
-npm install
-npm run build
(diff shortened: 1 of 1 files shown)


build/updater.sh
```diff
@@ -1,9 +1,9 @@
...
-cd query-builder
+cd /data/project/query-builder-test/query-builder
+git reset --hard origin/master
 git pull
-rm -rf node_modules
-npm install -g npm@6.14.8
-npm install
-npm run build
+./node_modules/.bin/npm install npm@6.14.8
+./node_modules/.bin/npm update
```
jakobw: [`npm update`](https://docs.npmjs.com/cli/update) is for updating packages. We want `npm install` or `npm ci`.

build/updater.sh
```diff
@@ -1,9 +1,9 @@
 #!/bin/bash
-cd query-builder
+cd /data/project/query-builder-test/query-builder
+git reset --hard origin/master
 git pull
-rm -rf node_modules
-npm install -g npm@6.14.8
-npm install
-npm run build
+./node_modules/.bin/npm install npm@6.14.8
```
jakobw: This cannot work the first time it runs, can it? There won't be an npm binary inside `node_modules/` so this should just use the system's npm for this step, right?!

jakobw (changes requested):
1 - 2 hours sounds really weird to me. We have setups for larger projects with the same tech stack and it takes a minute or two. Not sure what's going on there :/

build/updater.sh
Ladsgroup: Done

build/updater.sh
Ladsgroup: Done

jakobw (approved):
LGTM and haven't seen any further hiccups on the test system so yay!!

PR #21, opened 2020-10-06
Use CSS logical properties

jakobw:
This allows us to use [CSS Logical Properties](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Logical_Properties) in our SCSS code.

Also fixes a linting issue with 29b1c52 that caused certain TS eslint rules to be enabled in JS files.

- [x] add the necessary PostCSS plugins
- [x] enforce logical properties through stylelint

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -25,6 +25,17 @@ module.exports = {
 			env: {
 				jest: true
 			}
+		},
+		{
+			files: [ '*.js' ],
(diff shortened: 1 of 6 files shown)


jakobw:
This also fixes the broken paddings on the TextInputs in Chrome

Ladsgroup (approved):

PR #26, opened 2020-10-07
Generate SPARQL with hardcoded property

guergana:
* trigger vuex action on button click
* build sparql query based on the contents of the Vuex store
* log the query to the console from within the vuex action

Bug: T264321


src/store/mutations.ts
```diff
@@ -1,7 +1,50 @@
...
+import QueryBuilderSparqlGenerator from '@/sparql/QueryBuilderSparqlGenerator';
+import SparqlGenerator from '@/sparql/SparqlGenerator';
 
 export default {
 	setValue( state: RootState, value: string ): void {
 		state.value = value;
+	},
+	runQuery( state: RootState ): void {
+		const prefixes = {
+			wdt: 'http://www.wikidata.org/prop/direct/'
```
Ladsgroup: you can load all prefixes here by loading the allNamespaces variable. OTOH, I'm planning to move that file, so it's fine for now.

src/store/mutations.ts
```diff
@@ -1,7 +1,50 @@
...
+		};
+		const queryObject = {
```
Ladsgroup: This definitely needs a builder a dedicated types but that's for later, I made https://phabricator.wikimedia.org/T264898 and https://phabricator.wikimedia.org/T264897 for it.

Ladsgroup:
Tested locally and it works fine, it LGTM but someone who knows Vue better should double check.

jakobw (changes requested):
Omg, a first SPARQL query!! Awesome to see this working! :tada:

With this being a central piece in the application we're building, I think it's important to get these first steps right from the start to save us from trouble later. Looking at it now, I think we made a mistake during our task breakdown when deciding that the SPARQL query should be generated within a store action. This doesn't make any sense, because we're not writing anything into the store. We just want to read the store contents and generate a query.

<sub>(yes, I remember I made up and wrote the task description during our task breakdown. Sorry! :man_facepalming:  )</sub>

Instead, I think we should only *read* from the store when the "run query" button is clicked, and everything else should happen outside the store. Logging the query to the console can happen in a regular component method. Generating the query can be fully decoupled from vue/vuex which also makes it easier to _test_ and use proper type declarations.

Here is how I think we could separate the puzzle pieces:

1. create a `QueryRepresentation` (ideas for better names welcome) type which is the part of the store containing the bits and pieces that are required to generate the query, and also the input to the function that generates the sparql code. For now this type is identical to our `RootState` type, but it'll only be a subset of the state eventually.
    Something like this:
    ```typescript
    interface QueryRepresentation {
        property: string;
        value: string;
    }
    
    interface RootState extends QueryRepresentation {}
    ```
   or making the `QueryRepresentation` only a part of the `RootState` instead of extending it. The latter would involve small changes to our existing vuex code.
   
2. create a store getter returning the relevant part of the store with `QueryRepresentation` as its return type

3. move the sparql generating code out of the vuex mutation into a dedicated function (in the sparql/ directory?) `( query: QueryRepresentation ) => string` building the SPARQL query from the `QueryRepresentation`. This makes it testable in isolation without store/component binding which is nice. And we definitely want tests for this!

4. wire things up in the component. It could look something like
    ```typescript
    runQuery(): void {
        console.log( generateSparql(
            this.$store.getters.queryRepresentation
        ) );
    }
    ```

Also +1 to everything Amir said.

src/store/mutations.ts
guergana: Done. Please check 4d12ac1b5eb1b3bff8e1eeb1276fd94aee697499

guergana:
Paired on this PR with @jakobw

Ladsgroup:
LGTM, I don't know if Jakob's notes all are addressed or not. Waiting for him.

jakobw (approved):
Yes, looks good to me, too! :sparkles:

PR #27, opened 2020-10-07
Type sparqljs-related objects

Ladsgroup:
And enforcing it in SparqlGenerator

Bug: T264897


jakobw (changes requested):
It looks like someone already solved this problem for us: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/sparqljs/index.d.ts

This should be as easy as running `docker-compose run --rm node npm install @types/sparqljs --save-dev` and importing the respective types. That way we can also import library using the normal ES module syntax instead of using `require`.

Ladsgroup:
> It looks like someone already solved this problem for us: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/sparqljs/index.d.ts
> 
> This should be as easy as running `docker-compose run --rm node npm install @types/sparqljs --save-dev` and importing the respective types. That way we can also import library using the normal ES module syntax instead of using `require`.

Oh thanks. I tried it and you can see the result: https://github.com/wmde/query-builder/pull/28/checks?check_run_id=1228011077

It seems either the type is incompatible with the spraqljs we have or it's buggy :(

Ladsgroup:
Tried with an earlier version of the types library and still doesn't work.

jakobw (changes requested):
The `.d.ts` file for sparqljs looks good and I confirmed locally that it works! Yayy!!

I think we can go a step further and use it in the same way that the `@types/sparqljs` package would have worked, i.e. as a [module type definition](https://www.typescriptlang.org/docs/handbook/declaration-files/templates/module-d-ts.html) of the `sparqljs` module. This should be as easy as:

1. moving the .d.ts file to a more fitting place `src/sparql/Sparqljs.d.ts -> src/@types/sparqljs/index.d.ts`
2. adding this folder as a new type root in `tsconfig.json`:
    ```
    "typeRoots": [
        "node_modules/@types",
        "src/@types"
    ],
    ```
3. using the types as if they came from `sparqljs` directly
    ```
	import {
		Generator as SparqlGenerator,
		Parser as SparqlParser,
		SelectQuery
	} from 'sparqljs';
    ```
4. get rid of all `require( 'sparqljs' )` calls and replace with proper module imports as shown above

Let me know if that makes sense. I tried it locally for `QueryBuilderSparqlGenerator.spec.ts` and it seemed to do the trick.

Ladsgroup:
> The `.d.ts` file for sparqljs looks good and I confirmed locally that it works! Yayy!!
> 
> I think we can go a step further and use it in the same way that the `@types/sparqljs` package would have worked, i.e. as a [module type definition](https://www.typescriptlang.org/docs/handbook/declaration-files/templates/module-d-ts.html) of the `sparqljs` module. This should be as easy as:
> 
>     1. moving the .d.ts file to a more fitting place `src/sparql/Sparqljs.d.ts -> src/@types/sparqljs/index.d.ts`
> 
>     2. adding this folder as a new type root in `tsconfig.json`:
>        ```
>        "typeRoots": [
>            "node_modules/@types",
>            "src/@types"
>        ],
>        ```
> 
>     3. using the types as if they came from `sparqljs` directly
>        ```
>        import {
>        	Generator as SparqlGenerator,
>        	Parser as SparqlParser,
>        	SelectQuery
>        } from 'sparqljs';
>        ```
> 
>     4. get rid of all `require( 'sparqljs' )` calls and replace with proper module imports as shown above
> 
> 
> Let me know if that makes sense. I tried it locally for `QueryBuilderSparqlGenerator.spec.ts` and it seemed to do the trick.

Thanks. I tried it and it seems that it didn't work, Can you take a look to see what I'm doing wrong? I can't figure out what's wrong :/

tests/unit/sparql/QueryBuilderSparqlGenerator.spec.ts
```diff
@@ -1,11 +1,15 @@
 import QueryBuilderSparqlGenerator from '@/sparql/QueryBuilderSparqlGenerator';
-import SparqlGenerator from '@/sparql/SparqlGenerator';
-import Sparqljs from '@/sparql/Sparqljs';
+import {
+	Generator as SparqlGenerator,
+	Parser as SparqlParser,
+	SelectQuery,
+} from "@/@types/sparqljs";
```
Ladsgroup: just trying to import it from "sparqljs" didn't work in my localhost, I don't know why.

jakobw: ~~Worked for me. Uploaded fixes!~~ Nope, I broke it.

jakobw:
> Thanks. I tried it and it seems that it didn't work, Can you take a look to see what I'm doing wrong? I can't figure out what's wrong :/

blerrrgh, sorry @Ladsgroup , my instructions were incomplete and it only worked for me because I had a patched version of the type declarations still flying around in my node_modules/ folder. It was missing the `declare module 'sparqljs' {` in the .d.ts. Fixed it for realz this time. I removed the `generates empty string` test. I don't think we really want to allow that anyway.

jakobw (changes requested):
Looks good except that the title of the PR doesn't fit the contents anymore! :D

The changes to `package-lock.json` don't belong here, otherwise the code looks good to go.

Ladsgroup:
> Looks good except that the title of the PR doesn't fit the contents anymore! :D
> 
> The changes to `package-lock.json` don't belong here, otherwise the code looks good to go.

Thanks! Changed the title and removed the package-lock.json

jakobw (approved):
YAY, lgtm! Sorry again about the confusing "help".

Once we have an upstream PR it would be nice to link to it in a comment here.

Ladsgroup:
> YAY, lgtm! Sorry again about the confusing "help".

It was actually pretty helpful. Thanks.

> Once we have an upstream PR it would be nice to link to it in a comment here.

I will!

PR #29, opened 2020-10-09
Introduce QueryObjectBuilder

Ladsgroup:
Simple builder that takes a QueryRepresentation and produce the object
sparqljs would understand.

Currently, the types are not in master yet and as the result, lints fail
for using a general "any" type.

### DIFF
--- a/src/sparql/QueryObjectBuilder.ts
+++ b/src/sparql/QueryObjectBuilder.ts
@@ -1,8 +1,9 @@
-import allNamespaces from '@/sparql/RdfNamespaces';
+import rdfNamespaces from '@/sparql/rdfNamespaces';
+import QueryRepresentation from '@/sparql/QueryRepresentation';
+import { SelectQuery } from 'sparqljs';
 
 export default class QueryObjectBuilder {
-	// TODO: Change the type once #27 is merged
-	private queryObject: any;
+	private queryObject: SelectQuery;
 
 	public constructor() {
 		this.queryObject = {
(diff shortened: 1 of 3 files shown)


src/sparql/QueryObjectBuilder.ts
```diff
@@ -0,0 +1,48 @@
...
+		this.queryObject = {
+			queryType: 'SELECT',
+			variables: [],
+			where: [],
+			type: 'query',
+			prefixes: allNamespaces
+		};
+	}
+
+	public buildFromQueryRepresentation( queryRepresentation: any ) {
```
guergana: maybe this can be changed already? was checking the types and not sure which it corresponds to

guergana (changes requested):
update type pleasee

src/sparql/QueryObjectBuilder.ts
Ladsgroup: Done.

tests/unit/sparql/QueryObjectBuilder.spec.ts
```diff
@@ -0,0 +1,51 @@
+import allNamespaces from '@/sparql/rdfNamespaces';
+import QueryObjectBuilder from '@/sparql/QueryObjectBuilder';
+
+describe( 'QueryObjectBuilder', () => {
+	it( 'einfach', () => {
```
jakobw: <_<

jakobw (changes requested):
Looks good. The fact that QueryObjectBuilder isn't working with an immutable query object scares me a little. This could go wrong if we don't keep that in mind, but it's fine for now.

I'm also not 100% clear yet how we want to slice and dice the SPARQL generating code. We for sure want some sort of builder, but I was wondering whether we want the builder to be tied to our QueryRepresentation type, or something more generic. Hard to tell.

Needs a better test name, otherwise good to go from my side! :rocket:

Ladsgroup:
> Looks good. The fact that QueryObjectBuilder isn't working with an immutable query object scares me a little. This could go wrong if we don't keep that in mind, but it's fine for now.

I understand that mutable objects are scary but the current builder pattern is pretty common and is not that bad (as you said, if we keep it in mind). We can revisit and clean it up later if it turns out to be bad.
> 
> I'm also not 100% clear yet how we want to slice and dice the SPARQL generating code. We for sure want some sort of builder, but I was wondering whether we want the builder to be tied to our QueryRepresentation type, or something more generic. Hard to tell.

Same but also, AFAIK it's supposed to be "Simple" query builder so I hope nothing to complicated shows up at our doorstep.
> 
> Needs a better test name, otherwise good to go from my side! rocket

Damn, I was hoping this would goes unnoticed :D

jakobw (approved):
:cool: :cool: :cool:

PR #31, opened 2020-10-12
Display query result on iframe

bereket-WMDE:
Bug: T263568

### DIFF
--- a/src/components/QueryBuilder.vue
+++ b/src/components/QueryBuilder.vue
@@ -23,6 +23,21 @@
 		<div class="querybuilder__run">
 			<Button @click.native="runQuery" type="primaryProgressive">Run query</Button>
 		</div>
+		<div class="querybuilder__result">
+			<div v-if="encodedQuery.length === 0" class="querybuilder__result__description">
+				Results will be displayed here
+			</div>
+
+			<!-- TODO use global config variables instead of hardcoding the url-->
+			<iframe
+				v-else
+				:src="'https://query.wikidata.org/embed.html#' + encodedQuery"
+				:key="iframeRenderKey"
+				class="querybuilder__result__iframe"
+				referrerpolicy="origin"
+				sandbox="allow-scripts allow-same-origin">
+			</iframe>
+		</div>
 	</div>
 </template>
 
@@ -35,14 +50,21 @@ import buildQuery from '@/sparql/buildQuery';
 
 export default Vue.extend( {
 	name: 'QueryBuilder',
+	data() {
+		return {
+			encodedQuery: '',
+			iframeRenderKey: 0
+		};
+	},
 	methods: {
 		updateInputTextValue( value: string ): void {
 			this.$store.dispatch( 'updateValue', value );
 		},
 		runQuery(): void {
-			console.log( buildQuery(
-				this.$store.getters.query
-			) );
+			this.encodedQuery = encodeURI( buildQuery( this.$store.getters.query ) );
+
+			// force the iframe to rerender https://stackoverflow.com/a/48755228
+			this.iframeRenderKey = Math.random();
 		}
 	},
 	computed: {
@@ -102,5 +124,25 @@ export default Vue.extend( {
 
 .querybuilder__run {
 	margin-block-start: $dimension-layout-medium;
+}
+
+.querybuilder__result {
+	margin-block-start: $dimension-layout-small;
+}
+
+.querybuilder__result__description {
+	padding-block: $dimension-spacing-xxlarge;
+	font-size: $font-size-style-description;
+	font-family: $font-family-style-description;
+	color: $font-color-subtle;
+	font-weight: $font-weight-style-description;
+	line-height: $font-line-height-style-description;
+	text-align: center;
+}
+
+.querybuilder__result__iframe {
+	width: $dimension-width-full;
+	border: none;
+	height: 95vh;
 }
 </style>


guergana:
Looks great! but seems the styling is tbd still

bereket-WMDE:
> Looks great! but seems the styling is tbd still

yap. waiting on charlie

src/components/QueryBuilder.vue
```diff
@@ -35,14 +44,22 @@ import buildQuery from '@/sparql/buildQuery';
...
 	methods: {
 		updateInputTextValue( value: string ): void {
 			this.$store.dispatch( 'updateValue', value );
 		},
 		runQuery(): void {
-			console.log( buildQuery(
-				this.$store.getters.query
-			) );
+			this.showQueryResults = false;
+			Vue.nextTick( () => {
```
Ladsgroup: I'm not sure using Vue.nextTick is necessary

Ladsgroup:
Tested locally and it works just fine. This nitpick there and then I think it's good to go.

src/components/QueryBuilder.vue
bereket-WMDE: thats whehn if someone changes the value and clicks on run again.
then the iframe needs to be reloaded (with the new value ofc)

src/components/QueryBuilder.vue
```diff
@@ -23,6 +23,15 @@
 		<div class="querybuilder__run">
 			<Button @click.native="runQuery" type="primaryProgressive">Run query</Button>
 		</div>
+		<div class="querybuilder__result">
+			<div v-if="!showQueryResults" class="querybuilder__result__description">Results will be displayed here</div>
+			<iframe v-bind:src="'https://query.wikidata.org/embed.html#' + encodedQuery"
```
Ladsgroup: I assume this would be accessible to third party installations of Wikibase as well, so this should be configurable. You can file a bug or todo somewhere for this.

src/components/QueryBuilder.vue
```diff
@@ -23,6 +23,15 @@
...
+			class="querybuilder__result__iframe"
+			referrerpolicy="origin"
+			v-if="showQueryResults"
+			sandbox="allow-scripts allow-same-origin allow-popups">
```
Ladsgroup: Do we need "allow-popups"?

src/components/QueryBuilder.vue
Ladsgroup: you can use the shorthand ":" instead of "v-bind"

src/components/QueryBuilder.vue
Ladsgroup: oh I see I think. Thanks.

Ladsgroup: Maybe a comment explaining why? but that's a nitpick. I don't mind.

src/components/QueryBuilder.vue
bereket-WMDE: did you mean this: https://query.wikidata.org/embed.html should be configurable ?

Ladsgroup: Yes

bereket-WMDE: i can try and set up and application config object containing Vue application global configurations.
or were you thinking somewhere in the direction of enviroment variables ?

src/components/QueryBuilder.vue
bereket-WMDE: we don't. 
tested it and removed it. thanks for the tip

src/components/QueryBuilder.vue
bereket-WMDE: sure. will add a comment

src/components/QueryBuilder.vue
bereket-WMDE: Thanks for the tip

Ladsgroup: I was more thinking of the former (Vue global application configurations). If it grows too big, I'm fine with hard-coding it for now and doing it later (just tracking the tech debt somewhere, either a todo or a phabricator ticket)

Ladsgroup (approved):
LGTM

sai-san:
Hello there! Leaving a couple of quick comments here after checking against the designs: 

- There's a 24px vertical spacing (dimension-layout-small) missing between the "Run query" button and the results div.

- In the results' placeholder screen (before the query is run): the 32px internal padding applied to the div containing the results' description ("Results will be displayed here") should only be applied to the top and the bottom. 

- When results are loaded, the results div should not display the current 32px padding: this will make the iframe to align horizontally with the "Run query button".

bereket-WMDE:
> Hello there! Leaving a couple of quick comments here after checking against the designs:
> 
>     * There's a 24px vertical spacing (dimension-layout-small) missing between the "Run query" button and the results div.
> 
>     * In the results' placeholder screen (before the query is run): the 32px internal padding applied to the div containing the results' description ("Results will be displayed here") should only be applied to the top and the bottom.
> 
>     * When results are loaded, the results div should not display the current 32px padding: this will make the iframe to align horizontally with the "Run query button".


I have made the changes. can be verified again. 
feel free for feedback, if i miss something

sai-san:
Thanks for the changes, @bereket-WMDE! There's only one little thing missing: In the results' placeholder screen (before the query is run): the internal padding (dimension-spacing-xxlarge) applied to the div containing the results' description ("Results will be displayed here") should be applied to the bottom too. (This way, the area will be nicely spaced when the border is added).

After syncing with Charlie, we found that the current implementation is overlooking one of the specifications. The fact that the default height of the iframe should be adjusted to the size of its content (results table). As specified in the [Phab ticket](https://phabricator.wikimedia.org/T263568): "default height = default iframe height based on its content". At the moment, the max height (97vh) is being used as the default height.

This is something that @guergana had noticed and shared in our channel, so I'm pinging her here to clarify that, in fact, it was specified that the iframe's height should be adjusted whenever fewer results are displayed.

Thanks again! 🙏

PR #32, opened 2020-10-15
Add role attributes for accessibility

Ladsgroup:
These are required in ARIA a11y guidelines.
Some info:
https://dequeuniversity.com/rules/axe/3.5/landmark-one-main
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/heading_role

### DIFF
--- a/src/components/QueryBuilder.vue
+++ b/src/components/QueryBuilder.vue
@@ -1,6 +1,6 @@
 <template>
 	<div class="querybuilder" role="main">
-		<h1 class="querybuilder__heading" role="heading" aria-level="1">Simple Query Builder</h1>
+		<h1 class="querybuilder__heading">Simple Query Builder</h1>
 		<p class="querybuilder__description">
 			Short explanatory text that also manages expectations. Short explanatory text that also
 			manages expectations.
--- a/src/components/QueryBuilder.vue
+++ b/src/components/QueryBuilder.vue
@@ -1,27 +1,31 @@
 <template>
-	<div class="querybuilder">
-		<h1 class="querybuilder__heading">Simple Query Builder</h1>
-		<p class="querybuilder__description">
-			Short explanatory text that also manages expectations. Short explanatory text that also
-			manages expectations.
-		</p>
-		<h2 class="querybuilder__find-title">Find all items...</h2>
-		<div class="querybuilder__rule">
-			<TextInput
-				class="querybuilder__rule__property"
-				label="Property"
-				:value="property.label"
-				placeholder="Enter a property" />
-			<TextInput
-				class="querybuilder__rule__value"
-				label="Value"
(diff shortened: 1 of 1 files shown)


jakobw:
The changes to the headings don't look right to me. Even the article you linked says that the heading structure we had before is actually preferable:

> Instead of using a `<div>` or `<span>` with a heading role and aria-level, consider using a native `<h1>` through `<h6>` element instead to indicate that this text is a heading, and what part of the structure it represents.

-> https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/heading_role

That said, adding `role="main"` to the element around the main content, and adding `role="form"` around the element containing the actual query builder seem like the right thing to do.

Ladsgroup:
> The changes to the headings don't look right to me. Even the article you linked says that the heading structure we had before is actually preferable:
> 
> > Instead of using a `<div>` or `<span>` with a heading role and aria-level, consider using a native `<h1>` through `<h6>` element instead to indicate that this text is a heading, and what part of the structure it represents.
> 
> -> https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/heading_role
> 
> That said, adding `role="main"` to the element around the main content, and adding `role="form"` around the element containing the actual query builder seem like the right thing to do.

you're right, it's just so complicated (a landmark should not be inside another landmark, every tag should be covered by a landmark, etc.) that sometimes it makes tricky to work it out. With the change I just did, it passes and doesn't put heading on a <div> element. Hope that works now.

src/components/QueryBuilder.vue
```diff
@@ -1,27 +1,29 @@
 <template>
-	<div class="querybuilder">
-		<h1 class="querybuilder__heading">Simple Query Builder</h1>
+	<div class="querybuilder" role="main">
+		<h1 class="querybuilder__heading" role="heading" aria-level="1">Simple Query Builder</h1>
```
jakobw: I'm still not convinced. Doesn't https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/heading_role say that `<h1>` on its own is equivalent to another element with ` role="heading" aria-level="1"` set? Can't we just leave it as `<h1 class="querybuilder__heading">`?

Ladsgroup: oh that part, I read somewhere it's better to keep html5 and ARIA identifiers.
https://dequeuniversity.com/rules/axe/3.5/landmark-one-main
> It is a best practice to use both HTML 5 and ARIA landmarks to ensure all content is contained within a navigational region.

But OTOH, this is not a landmark. I'd remove the heading bit.

jakobw (approved):
LGTM, thanks!! :sparkles:

PR #33, opened 2020-10-16
Add jest-axe to test basic a11y

Ladsgroup:
Bug: T265299

### DIFF
--- a/tests/unit/components/QueryBuilder.spec.ts
+++ b/tests/unit/components/QueryBuilder.spec.ts
@@ -1,6 +1,7 @@
 import { TextInput } from '@wmde/wikit-vue-components';
 import Vuex, { Store } from 'vuex';
-import { shallowMount, createLocalVue } from '@vue/test-utils';
+import { shallowMount, createLocalVue, mount } from '@vue/test-utils';
+import { axe, toHaveNoViolations } from 'jest-axe';
 import QueryBuilder from '@/components/QueryBuilder.vue';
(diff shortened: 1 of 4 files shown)


Ladsgroup:
Rebasing this on master, it should pass now.

micgro42:
Fixed the linting issues to showcase that it can actually get green. For the real implementation, we probably want it in a different place, i.e. in its own test job and not mixed into the unit tests.

bereket-WMDE (approved):

Ladsgroup:
The test is now failing for an actual a11y issue: https://phabricator.wikimedia.org/T265245#6637658

tests/unit/components/QueryBuilder.spec.ts
```diff
@@ -111,4 +111,30 @@ describe( 'QueryBuilder.vue', () => {
 		} );
 
 	} );
+
+	it( 'should not have obvious accessibility issues', async () => {
```
micgro42: could we move this out of the unit tests and into its own `a11y` test directory that is run on `npm run test:a11y`?

We might want to add more test scenarios in the future.

Ladsgroup: Done.

micgro42 (approved):

Ladsgroup:
Manual rebase ^

Ladsgroup:
The failure is unrelated and will be fixed in #83

PR #34, opened 2020-10-26
Enable `allow-popups` for the query results iframe

jakobw:
This is needed to let links in the results table open in a new window
when clicked.

Bug: T263568


Ladsgroup (approved):
It reduces the security but it's needed so let's do it.

PR #35, opened 2020-10-28
Add SearchEntityRepository interface

micgro42:
~~We probably do not want to use pagination in any of the use cases for
this hike as it would require multiple requests and that is probably a
bad experience for a autocomplete.~~
Turns out we are going to need pagination after all.

The actual server response contains a lot more information than just
what is currently defined in SearchResult. But it seems sensible to keep
the interface limited to what we really need just for now.

Bug: T266555

### DIFF
--- a/src/data-access/SearchEntityRepository.ts
+++ b/src/data-access/SearchEntityRepository.ts
@@ -1,0 +1,9 @@
+import SearchResult from './SearchResult';
+
+/**
+ * Repository to search for entities.
+ * The language will be defined in the constructor as will be further options
+ */
+export default interface SearchEntityRepository {
+    searchProperties( searchString: string, limit?: number, offset?: number ): Promise<Array<SearchResult>>;
+}
--- a/src/data-access/SearchResult.ts
+++ b/src/data-access/SearchResult.ts
@@ -1,0 +1,5 @@
+export default interface SearchResult {
+	id: string,
+	label: string,
+	description: string,
+}


guergana (dismissed):
LGTM!

micgro42:
This needs adjustments for [T266856](https://phabricator.wikimedia.org/T266856)

Ladsgroup:
> This needs adjustments for [T266856](https://phabricator.wikimedia.org/T266856)

Do you think this necessary needs to happen now or we can do it in small batch and adjust it later?

I also think the file should end with .d.ts (as it's a type and not much actual code)

guergana:
> > This needs adjustments for [T266856](https://phabricator.wikimedia.org/T266856)
> 
> Do you think this necessary needs to happen now or we can do it in small batch and adjust it later?
> 
> I also think the file should end with .d.ts (as it's a type and not much actual code)

Agree with making it a `.d.ts` file

micgro42:
> > This needs adjustments for [T266856](https://phabricator.wikimedia.org/T266856)
> 
> Do you think this necessary needs to happen now or we can do it in small batch and adjust it later?

It makes sense to me to have the interface look like it is supposed to look right away. It is just adding another parameter `offset?: number` or something similar.

> I also think the file should end with .d.ts (as it's a type and not much actual code)

We actually discussed that [yesterday](https://mattermost.wikimedia.de/swe/pl/p9xfgg54c38ndkbqaxpimu6ksc) quite a bit. I don't think `.d.ts` is the right ending, because it is specifically intended to provide ambient typing for pure JavaScript libraries. And that is not what this is. This is the equivalent to a PHP interface.

There was the suggestion to name that file `.type.ts` or something similar. I'm more open to that and added it to the upcoming retro to be discussed as it also has downsides.

Ladsgroup (approved):

PR #36, opened 2020-10-30
Add implementation for repository to search for entities

micgro42:
Bug: T266549

### DIFF
--- a/src/data-access/FetchSearchEntityRepository.ts
+++ b/src/data-access/FetchSearchEntityRepository.ts
@@ -1,0 +1,52 @@
+import SearchEntityRepository from '@/data-access/SearchEntityRepository';
+import SearchResult from '@/data-access/SearchResult';
+import TechnicalProblem from '@/data-access/errors/TechnicalProblem';
+
+export default class FetchSearchEntityRepository implements SearchEntityRepository {
+	private readonly forLanguageCode: string;
+	private readonly endpoint: string;
+
+	public constructor( languageCode: string, endpoint: string ) {
+		this.forLanguageCode = languageCode;
+		this.endpoint = endpoint;
+	}
+
+	public async searchProperties( searchString: string, limit?: number, offset?: number ):
+	Promise<SearchResult[]> {
+		const params: { [key: string]: string } = {
+			action: 'wbsearchentities',
+			search: searchString,
+			language: this.forLanguageCode,
+			type: 'property',
+			format: 'json',
+			formatversion: '2',
+			errorformat: 'plaintext',
+			origin: '*'
+		};
+		if ( limit ) {
+			params.limit = `${limit}`;
+		}
+		if ( offset ) {
+			params.offset = `${offset}`;
+		}
+
+		const url = new URL(this.endpoint);
+		for ( const key in params ) {
+			url.searchParams.set( key, params[key] );
+		}
+		let response: Response;
+		try {
+			response = await fetch( url.toString() );
+		} catch ( e ) {
+			throw new TechnicalProblem( 'Network error' );
+		}
+
+		if ( !response.ok ) {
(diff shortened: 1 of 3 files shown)


src/data-access/FetchSearchEntityRepository.ts
```diff
@@ -0,0 +1,48 @@
...
+			params.limit = limit;
+		}
+		if ( offset ) {
+			params.offset = offset;
+		}
+
+		const query = Object.entries( params )
+			.map( ( entry ) => entry.join( '=' ) )
+			.join( '&' );
+		const url = `${this.endpoint}?${query}`;
```
Ladsgroup: This will be prone to reflective XSS attacks. You can use URL object https://javascript.info/url instead.

micgro42: How? This URL never makes its way into HTML.

But you are probably right, using the URL objects should result in nicer code. I'll look for an [IE polyfill](https://caniuse.com/?search=URL)

Ladsgroup: > How? This URL never makes its way into HTML.
It can break out and run custom javascript here.
> 
> But you are probably right, using the URL objects should result in nicer code. I'll look for an [IE polyfill](https://caniuse.com/?search=URL)
Half of the things we use is not available in IE11 (most of ES6), I thought we do babbel for that (We should!)

micgro42: The polyfill seems to come with core-js, so we should be good. :heavy_check_mark:

micgro42: >> How? This URL never makes its way into HTML.
>
> It can break out and run custom javascript here.

Still not seeing how. Could you explain? This url is used in a fetch call. The only way that I see, that this could maybe used for XSS would be if we were to display the unescaped URL in an error message.

src/data-access/FetchSearchEntityRepository.ts
```diff
@@ -0,0 +1,52 @@
...
+		} catch ( e ) {
+			throw new TechnicalProblem( 'Network error' );
+		}
+
+		if ( !response.ok ) {
+			throw new TechnicalProblem( `${response.status}: ${response.statusText}` );
+		}
+
+		const data = await response.json();
+
```
micgro42, line 49: I wonder if we should add error handling in case there are warnings? Though afaik, we do not have client side error logging, so this is probably a bit moot.

src/data-access/FetchSearchEntityRepository.ts
```diff
@@ -0,0 +1,52 @@
...
+			params.offset = `${offset}`;
+		}
+
+		const url = new URL(this.endpoint);
+		for ( const key in params ) {
+			url.searchParams.set( key, params[key] );
+		}
+		let response: Response;
+		try {
+			response = await fetch( url.toString() );
```
micgro42, line 39: I guess this will add the query builder's URL as the referrer? Should we somehow additionally indicate that these requests come from the QueryBuilder for tracking purposes?

src/data-access/FetchSearchEntityRepository.ts
Ladsgroup: What I meant was when you concat user input string, it can break out and run arbitrary javascript. I can find some materials in the internet.

src/data-access/FetchSearchEntityRepository.ts
Ladsgroup: Browser will add the referrer, we don't need to do anything here.

src/data-access/FetchSearchEntityRepository.ts
Ladsgroup: Yeah. I don't see the point.

Ladsgroup (approved):
LGTM

PR #37, opened 2020-11-02
Wire up the property to the store

Ladsgroup:
This will be used in the next iterations for dynamic property

Bug: T266552

### DIFF
--- a/src/data-access/SearchEntityRepository.ts
+++ b/src/data-access/SearchEntityRepository.ts
@@ -5,5 +5,5 @@ import SearchResult from './SearchResult';
  * The language will be defined in the constructor as will be further options
  */
 export default interface SearchEntityRepository {
-    searchProperties( searchString: string, limit?: number, offset?: number ): Promise<Array<SearchResult>>;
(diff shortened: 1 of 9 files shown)


micgro42:
Thanks! Looks good so far. I just now realized that we need more or less the same also for the property search results that we are going to feed to the property lookup.

Ladsgroup:
> Thanks! Looks good so far. I just now realized that we need more or less the same also for the property search results that we are going to feed to the property lookup.

yeah, we slowly should move towards modules (and namespaces) but that can happen later. Do you want the search results be done in this PR?

micgro42:
> > Thanks! Looks good so far. I just now realized that we need more or less the same also for the property search results that we are going to feed to the property lookup.
> 
> yeah, we slowly should move towards modules (and namespaces) but that can happen later. Do you want the search results be done in this PR?

Not necessarily, though it might affect the property interface. But a second PR is probably easier to review as it is a second idea.

micgro42 (approved):

PR #38, opened 2020-11-02
Spin off the result part of query builder component to a child component

Ladsgroup:
Having "fat component" is an anti-pattern in Vue and it's highly
recommended to keep logic in *.vue files below some threshold.

This will also help in future when adding the error handling.

### DIFF
--- a/src/data-access/SearchEntityRepository.ts
+++ b/src/data-access/SearchEntityRepository.ts
@@ -5,5 +5,5 @@ import SearchResult from './SearchResult';
  * The language will be defined in the constructor as will be further options
  */
 export default interface SearchEntityRepository {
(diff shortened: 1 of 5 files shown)


src/data-access/SearchEntityRepository.ts
```diff
@@ -5,5 +5,5 @@ import SearchResult from './SearchResult';
  * The language will be defined in the constructor as will be further options
  */
 export default interface SearchEntityRepository {
-    searchProperties( searchString: string, limit?: number, offset?: number ): Promise<Array<SearchResult>>;
+	searchProperties( searchString: string, limit?: number, offset?: number ): Promise<SearchResult[]>;
```
micgro42, line 8: Those are duplicated with the other PR.

(Also, I'm wondering why our CI didn't catch that before it was merged.)

micgro42 (approved):

PR #39, opened 2020-11-02
Create service container to wire up the repository

micgro42:
This is heavily inspired by what was done in Termbox:
* https://github.com/wikimedia/wikibase-termbox/blob/4cc5b4ed6cbc2e58b9650bd01c80512e96a9c715/src/common/TermboxServices.ts
* https://github.com/wikimedia/wikibase-termbox/blob/4cc5b4ed6cbc2e58b9650bd01c80512e96a9c715/src/client-entry.ts
* https://github.com/wikimedia/wikibase-termbox/blob/4cc5b4ed6cbc2e58b9650bd01c80512e96a9c715/src/store/index.ts#L22

The repository will be actually used in #43

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -12,7 +12,7 @@ export default Vue.extend( {
 	name: 'App',
 	components: {
 		QueryBuilder
-	},
+	}
 } );
 </script>
 
--- a/src/App.vue
+++ b/src/App.vue
@@ -7,14 +7,12 @@
 <script lang="ts">
 import Vue from 'vue';
 import QueryBuilder from '@/components/QueryBuilder.vue';
-import store from '@/store';
 
 export default Vue.extend( {
 	name: 'App',
 	components: {
 		QueryBuilder
 	},
-	store
 } );
 </script>
 
--- a/src/QueryBuilderServices.ts
+++ b/src/QueryBuilderServices.ts
@@ -1,0 +1,25 @@
+import SearchEntityRepository from '@/data-access/SearchEntityRepository';
+
+export interface Services {
+	searchEntityRepository: SearchEntityRepository;
+}
+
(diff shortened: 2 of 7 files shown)


src/App.vue
```diff
@@ -7,14 +7,12 @@
...
 import QueryBuilder from '@/components/QueryBuilder.vue';
-import store from '@/store';
 
 export default Vue.extend( {
 	name: 'App',
 	components: {
 		QueryBuilder
-	},
-	store
+	}
```
jakobw, line 15: Whoops, well spotted.

jakobw (approved):
:rocket: looks good to me!

Correction on the "inspired by termbox" part: most of this code is coming from data bridge originally. It only landed in termbox with [this commit](https://github.com/wikimedia/wikibase-termbox/commit/40596c78a673351b4d3490ab884e32bc63d83e4b) after the hike was already over. Termbox had a service container of sorts, but it was a more naive implementation, and it was imported into the store as a module, not properly injected :)

src/App.vue
micgro42: Less "spotted", more "stumbled over"^^

micgro42:
> rocket looks good to me!
> 
> Correction on the "inspired by termbox" part: most of this code is coming from data bridge originally. It only landed in termbox with [this commit](https://github.com/wikimedia/wikibase-termbox/commit/40596c78a673351b4d3490ab884e32bc63d83e4b) after the hike was already over. Termbox had a service container of sorts, but it was a more naive implementation, and it was imported into the store as a module, not properly injected :)

Ah, nice piece of history! During bridge, it was always Termbox that was the example to be copied and improved upon. Nice to see that some things made their way back :)

PR #41, opened 2020-11-03
Add styles for links

Ladsgroup:
We added links to descriptions and now we need to add styles for that.

Based on https://www.figma.com/file/gi3JYwSUQYhKS0mHwK25cQwi/WiKit?node-id=2510%3A3550
and tokens

### DIFF
--- a/src/components/QueryBuilder.vue
+++ b/src/components/QueryBuilder.vue
@@ -75,6 +75,7 @@ export default Vue.extend( {
 </script>
 
 <style scoped lang="scss">
+<!-- TODO replace with link component once available-->
 a {
 	font-family: $font-family-style-link;
 	font-style: $font-size-style-link;
--- a/src/components/QueryBuilder.vue
+++ b/src/components/QueryBuilder.vue
@@ -75,7 +75,7 @@ export default Vue.extend( {
 </script>
 
 <style scoped lang="scss">
-# TODO replace with link component once available
+// TODO replace with link component once available
 a {
 	font-family: $font-family-style-link;
 	font-style: $font-size-style-link;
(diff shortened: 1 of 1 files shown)


sai-san:
Just checked locally and it looks good. Thanks, Amir! 🙏  I left a tiny reminder inline. Let me know what you think

src/components/QueryBuilder.vue
```diff
@@ -75,6 +75,19 @@ export default Vue.extend( {
 </script>
 
 <style scoped lang="scss">
+a {
```
sai-san, line 79:
```suggestion
<!-- TODO replace with link component once available-->
a {
```

Ladsgroup:
> Just checked locally and it looks good. Thanks, Amir! pray I left a tiny reminder inline. Let me know what you think

Done. Thanks!

src/components/QueryBuilder.vue
Ladsgroup: This actually fails because this is not valid css :(

src/components/QueryBuilder.vue
```diff
@@ -75,6 +75,20 @@ export default Vue.extend( {
 </script>
 
 <style scoped lang="scss">
+<!-- TODO replace with link component once available-->
```
micgro42: Needs to be an CSS comment, not an HTML comment. I failed to spot that earlier

```suggestion
// TODO replace with link component once available
```

micgro42 (changes requested):

src/components/QueryBuilder.vue
sai-san: agh! of course. sorry about that!

micgro42 (approved):

PR #42, opened 2020-11-03
Start error reporting

Ladsgroup:
This is basic error reporting, the message component is not yet released
so it can't be used but most of the work of showing the user is done
now.

Bug: T266788
Bug: T266787

### DIFF
--- a/src/components/QueryBuilder.vue
+++ b/src/components/QueryBuilder.vue
@@ -51,6 +51,8 @@ export default Vue.extend( {
 	},
 	methods: {
 		validate(): boolean {
+			this.errors = [];
+			this.$store.dispatch( 'setErrors', [] );
 			if ( !this.selectedItem && !this.textInputValue ) {
 				this.errors.push( {
 					// eslint-disable-next-line max-len
--- a/src/components/QueryBuilder.vue
+++ b/src/components/QueryBuilder.vue
@@ -39,6 +39,7 @@ import Button from '@wmde/wikit-vue-components/src/components/Button.vue';
 import QueryResult from '@/components/QueryResult.vue';
 import buildQuery from '@/sparql/buildQuery';
 import Property from '@/data-model/Property';
+import { mapState } from 'vuex';
 
 export default Vue.extend( {
 	name: 'QueryBuilder',
@@ -49,7 +50,30 @@ export default Vue.extend( {
 		};
 	},
 	methods: {
+		validate(): boolean {
+			if ( !this.selectedItem && !this.textInputValue ) {
+				this.errors.push( {
+					// eslint-disable-next-line max-len
+					message: 'Looks like the Query Builder was empty, please enter a valid query first, then try running it again',
+					type: 'notice',
+				} );
+				return false;
+			}
+
+			if ( !this.property || !this.textInputValue ) {
+				this.errors.push( {
+					message: 'One or more fields are empty. Please complete the query or select a fitting field type.',
+					type: 'error',
+				} );
+				return false;
+			}
+			return true;
+		},
 		runQuery(): void {
+			const isValid = this.validate();
+			if ( !isValid ) {
+				return;
+			}
 			this.encodedQuery = encodeURI( buildQuery( this.$store.getters.query ) );
 
 			// force the iframe to rerender https://stackoverflow.com/a/48755228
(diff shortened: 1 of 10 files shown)


Ladsgroup:
Tests needed

Ladsgroup:
> Tests needed

Test added.

micgro42:
:thinking: This adds the store infrastructure, but is not actually using it? Or am I missing something?

Ladsgroup:
> thinking This adds the store infrastructure, but is not actually using it? Or am I missing something?

It does. It's a computed property passed around using mapState(). You can check out the PR and try it yourself (just don't write anything and click on Run query)

micgro42:
> > thinking This adds the store infrastructure, but is not actually using it? Or am I missing something?
> 
> It does. It's a computed property passed around using mapState(). You can check out the PR and try it yourself (just don't write anything and click on Run query)

This is very strange magic. That helper just plainly assumes that `setErrors` actions and such exist? :upside_down_face: 

But it works :shrug:

micgro42:
That being said, this still needs more work for setting the validation/error states for each input field.

Ladsgroup:
> > > thinking This adds the store infrastructure, but is not actually using it? Or am I missing something?
> > 
> > 
> > It does. It's a computed property passed around using mapState(). You can check out the PR and try it yourself (just don't write anything and click on Run query)
> 
> This is very strange magic. That helper just plainly assumes that `setErrors` actions and such exist? upside_down_face
> 
> But it works shrug

The validate method dispatches setErrors: `this.$store.dispatch( 'setErrors', [] );`


> That being said, this still needs more work for setting the validation/error states for each input field.

I think that's different in nature and needs to be done in a follow up (I will do it, just don't want to make this PR super big)

micgro42:
Hey, with the recently merged PRs, some nasty merge conflicts were introduced. I rebased your PR (since I also broke it :p). Please check that everything is still alright

micgro42:
> > thinking This adds the store infrastructure, but is not actually using it? Or am I missing something?
> 
> It does. It's a computed property passed around using mapState(). You can check out the PR and try it yourself (just don't write anything and click on Run query)

Mh, but that could do just as well without the store and only with a local state called `error`, since the validation is done in a method and not an action. Though, I think it works for this pull request. We can move it when dealing with setting the error states on the individual inputs in the next PR.

src/components/QueryBuilder.vue
```diff
@@ -49,7 +50,32 @@ export default Vue.extend( {
 		};
 	},
 	methods: {
+		validate(): boolean {
+			this.errors = [];
+			this.$store.dispatch( 'setErrors', [] );
```
micgro42, line 55:
```ts
this.errors = [];
this.$store.dispatch( 'setErrors', [] );
```
Those two lines together are strange. Since we are mapping the state to `errors`, shouldn't be one enough?

Ladsgroup:
> > > thinking This adds the store infrastructure, but is not actually using it? Or am I missing something?
> > 
> > 
> > It does. It's a computed property passed around using mapState(). You can check out the PR and try it yourself (just don't write anything and click on Run query)
> 
> Mh, but that could do just as well without the store and only with a local state called `error`, since the validation is done in a method and not an action. Though, I think it works for this pull request. We can move it when dealing with setting the error states on the individual inputs in the next PR.

My reasoning is that with this app getting bigger and bigger (and the need to introduce more components), we certainly need the error state into Vuex

src/components/QueryBuilder.vue
Ladsgroup: IIRC, this is an issue with proxies in Vue2 (setting arrays) and can be removed in Vue3.

micgro42: We should look into migrating to Vue3. Probably, better sooner than later.

micgro42:
> > > > thinking This adds the store infrastructure, but is not actually using it? Or am I missing something?
> > > 
> > > 
> > > It does. It's a computed property passed around using mapState(). You can check out the PR and try it yourself (just don't write anything and click on Run query)
> > 
> > 
> > Mh, but that could do just as well without the store and only with a local state called `error`, since the validation is done in a method and not an action. Though, I think it works for this pull request. We can move it when dealing with setting the error states on the individual inputs in the next PR.
> 
> My reasoning is that with this app getting bigger and bigger (and the need to introduce more components), we certainly need the error state into Vuex

Yeah, and that is probably true for the actual validation code as well, that should move into an action. But we can do that later.

micgro42 (approved):

src/components/QueryBuilder.vue
Ladsgroup: I think it was a deliberate choice to go with vue2, I think it was because of consistency across the whole wmde codebase (and we can't migrate mediawiki micron frontends yet). Jakob knows the details.

PR #43, opened 2020-11-05
Use Lookup component

micgro42:
Working, but still a lot to do

Todo:
- **tests**
  - [x] tests for QueryBuilder
  - [x] tests for PropertyLookup
  - [x] tests for the new action
- ~~[ ] Add debounce to request for property search results~~ Will be done in extra PR/Story
- [x] double check that all the types make sense, esp. MenuItem and SearchResult
- [x] split the change to the mutation test into an extra commit (maybe other PR)
- [x] squash the other PropertyLookup commit into the preceding one

PR #44, opened 2020-11-05
Restructure state to make it more extendable

micgro42:
"condition" is the domain name for one clause or group in the query builder. The state for a condition will in the future both contain the data needed for the query itself and some metadata like the search results or the errors.

### DIFF
--- a/src/components/QueryBuilder.vue
+++ b/src/components/QueryBuilder.vue
@@ -12,14 +12,13 @@
 				<TextInput
 					class="querybuilder__rule__property"
 					label="Property"
-					:value="property.label"
+					:value="selectedItem.label"
 					placeholder="Enter a property" />
 				<TextInput
 					class="querybuilder__rule__value"
 					label="Value"
 					ref="value"
-					:value="textInputValue"
-					@input="updateInputTextValue"
+					v-model="textInputValue"
 					placeholder="Enter a value" />
 			</div>
 			<div class="querybuilder__run">
@@ -37,9 +36,9 @@
 import Vue from 'vue';
 import TextInput from '@wmde/wikit-vue-components/src/components/TextInput.vue';
 import Button from '@wmde/wikit-vue-components/src/components/Button.vue';
-import { mapState } from 'vuex';
 import QueryResult from '@/components/QueryResult.vue';
 import buildQuery from '@/sparql/buildQuery';
+import Property from '@/data-model/Property';
 
 export default Vue.extend( {
 	name: 'QueryBuilder',
@@ -50,9 +49,6 @@ export default Vue.extend( {
 		};
 	},
 	methods: {
-		updateInputTextValue( value: string ): void {
-			this.$store.dispatch( 'updateValue', value );
-		},
 		runQuery(): void {
 			this.encodedQuery = encodeURI( buildQuery( this.$store.getters.query ) );
 
@@ -61,10 +57,13 @@ export default Vue.extend( {
 		},
 	},
 	computed: {
-		...mapState( {
-			property: 'property',
-			textInputValue: 'value',
-		} ),
+		selectedItem: {
+			get(): Property { return this.$store.getters.property; },
+		},
+		textInputValue: {
+			get(): string { return this.$store.getters.value; },
+			set( value: string ): void { this.$store.dispatch( 'updateValue', value ); },
+		},
 	},
 	components: {
 		Button,
--- a/src/store/RootState.ts
+++ b/src/store/RootState.ts
@@ -1,4 +1,13 @@
-import QueryRepresentation from '@/sparql/QueryRepresentation';
+export default interface RootState {
+	conditionRow: ConditionRow;
+}
 
-export interface RootState extends QueryRepresentation {
+export interface ConditionRow {
+	propertyData: {
+		id: string;
+		label: string;
+	};
+	valueData: {
+		value: string;
+	};
 }
--- a/src/store/getters.ts
+++ b/src/store/getters.ts
@@ -1,12 +1,20 @@
-import { RootState } from './RootState';
+import RootState from './RootState';
 import QueryRepresentation from '@/sparql/QueryRepresentation';
 import Property from '@/data-model/Property';
 
 export default {
-	query( { property, value }: RootState ): QueryRepresentation {
-		return { property, value };
+	query( rootState: RootState ): QueryRepresentation {
+		return {
+			condition: {
+				propertyId: rootState.conditionRow.propertyData.id,
+				value: rootState.conditionRow.valueData.value,
+			},
+		};
 	},
-	getProperty( rootState: RootState ): Property {
-		return rootState.property;
+	property( rootState: RootState ): Property {
+		return rootState.conditionRow.propertyData;
+	},
+	value( rootState: RootState ): string {
+		return rootState.conditionRow.valueData.value;
 	},
 };
--- a/src/sparql/QueryObjectBuilder.ts
+++ b/src/sparql/QueryObjectBuilder.ts
@@ -34,11 +34,11 @@ export default class QueryObjectBuilder {
 						},
 						predicate: {
 							termType: 'NamedNode',
-							value: rdfNamespaces.wdt + queryRepresentation.property.id,
+							value: rdfNamespaces.wdt + queryRepresentation.condition.propertyId,
 						},
 						object: {
 							termType: 'Literal',
-							value: queryRepresentation.value,
+							value: queryRepresentation.condition.value,
 						},
 					},
 				],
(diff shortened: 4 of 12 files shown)


src/store/getters.ts
```diff
@@ -3,10 +3,13 @@ import QueryRepresentation from '@/sparql/QueryRepresentation';
...
-	query( { property, value }: RootState ): QueryRepresentation {
-		return { property, value };
+	query( rootState: RootState ): QueryRepresentation {
+		return rootState.condition.data;
 	},
 	getProperty( rootState: RootState ): Property {
-		return rootState.property;
+		return rootState.condition.data.property;
+	},
+	getValue( rootState: RootState ): string {
```
jakobw: TL;DR reconsider getter naming or to change the `query` getter to make it consistent.

I tripped a little bit over these getters following method naming conventions. If they're intended to be used like properties with [property-style access](https://vuex.vuejs.org/guide/getters.html#property-style-access), naming them as such seems clearer to me.

This inconsistency existed before this PR so this doesn't need to be addressed here.

src/components/QueryBuilder.vue
```diff
@@ -61,10 +57,13 @@ export default Vue.extend( {
...
 	},
 	computed: {
-		...mapState( {
-			property: 'property',
-			textInputValue: 'value'
-		} )
+		selectedItem: {
+			get(): Property { return this.$store.getters.getProperty; }
+		},
+		textInputValue: {
```
jakobw, line 63: This definitely looks neater than before :ok_hand:

src/store/RootState.ts
```diff
@@ -1,4 +1,7 @@
 import QueryRepresentation from '@/sparql/QueryRepresentation';
 
-export interface RootState extends QueryRepresentation {
+export interface RootState {
+	condition: {
+		data: QueryRepresentation;
```
jakobw: Is the extra `data` nesting level needed?

src/store/RootState.ts
```diff
@@ -1,4 +1,7 @@
 import QueryRepresentation from '@/sparql/QueryRepresentation';
 
-export interface RootState extends QueryRepresentation {
+export interface RootState {
```
jakobw: I imagined the `QueryRepresentation` type to be the container for all conditions that in combination form the query. The idea was for it to evolve as we add new query building features. Ideally, it would both represent the visual state of the query builder, and also be consumed by the code that builds the sparql query. I don't know if this incremental approach actually makes sense for domain objects. In this case it looks like it was confusing :/

Eventually, I think it could look something like this:
```typescript
interface QueryRepresentation {
	conditions: ConditionGroup;
}

type ConditionGroup = (Union | Minus)[];

type Union = {
	type: 'union';
	conditions: Condition[];
}

type Minus = {
	type: 'minus';
	conditions: Condition[];
}

type Condition = {
// property, value, ...
}
```
(not sure if that actually makes sense, and it should use better domain terms)

Long story short, I think `RootState` should stay as it is, and `condition` should move into `QueryRepresentation`.

Edit: This model doesn't make sense because we don't actually have what I called `Minus` here, but I think we still need two kinds of condition groups for "and" groups and "or" groups.

jakobw (dismissed):
Grouping `property` and `value` within a `condition` object definitely makes sense. Props for checking with UX and product to align on the term! :+1:

Added a comment for where I expected this addition to happen, but otherwise it looks good!

src/store/RootState.ts
micgro42: The reasoning for not doing this is #43
I'm torn here:
On the one hand, I would like, as you suggest, to have a pure representation of the domain logic in the state.
On the other hand, almost every bit of that logic comes with extra information that needs to also be represented. Like entity search results or errors for a specific component, etc. This becomes more complicate by considering whether the label of an entity is _really_ part of the query?

Merging those two feels wrong, having two trees with the same structure feels wrong as well.

src/store/getters.ts
micgro42: Done in an extra commit :)

src/store/RootState.ts
jakobw: You definitely have a point, and yes, I agree that the Property's label probably shouldn't have made it into the `QueryRepresentation` type. How about leaving the two completely separate, and putting a `buildQueryRepresentationFromStoreState( state: RootState ): QueryRepresentation` step in between? There would be some redundancy, but mainly in the type declarations.

Off-topic: Interesting, I don't know if I agree with the search results making a round trip through the store. Couldn't the results stay local to a PropertyLookup/ItemLookup component that wraps `Lookup` and encapsulates the request logic?

src/store/getters.ts
jakobw: Nice :tada:

src/store/RootState.ts
micgro42: > You definitely have a point, and yes, I agree that the Property's label probably shouldn't have made it into the `QueryRepresentation` type. How about leaving the two completely separate, and putting a `buildQueryRepresentationFromStoreState( state: RootState ): QueryRepresentation` step in between? There would be some redundancy, but mainly in the type declarations.

Mh, yes that makes sense. Eventually, that will need to be a (lossy) two-way transformation between Store and QueryRepresentation.

From that QueryRepresentation we have then a one-way transformation into SPARQL and another two-way transformation into JSON(?) for encoding it in the URL.

I'll see if I can adjust the naming.

micgro42: > Off-topic: Interesting, I don't know if I agree with the search results making a round trip through the store. Couldn't the results stay local to a PropertyLookup/ItemLookup component that wraps `Lookup` and encapsulates the request logic?

Mh, I'm torn again.
- On the one hand it would make a lot of things easier to encapsulate all that into a single component and not having to worry about the action writing the search results into the correct place in the store and the getter getting it from the correct place.
- On the other hand it feels strange to have the logic of initiating the request in a vue component, i.e. the presentation layer, and not more in the business logic layer like an action.

:thinking: 

But wrapping that DS Lookup in a PropertyLookup component is certainly a good start. Maybe this will be easier to judge once that component is there.

micgro42:
I think in the long run we could end up with something like:

```ts
export default interface RootState {
	conditionRows: ConditionRow[];
	errors: Error[];
}

export interface ConditionRow {
	propertyData: {
		id: string;
		label: string;
		hasError?: false|'error'|'warning'; // ignored for now
		errorMessage?: string; // ignored for now
	};
	valueData: {
		value: string; // probably going to be a couple of different types depending on string/item/quantity/date ...
		valueType?: string; // ignored for now, probably an enum: unknown|string|item|quantity|date|...
		hasError?:  false|'error'|'warning'; // ignored for now
		errorMessage?: string; // ignored for now
	}
	negated?: boolean; // ignored for now
	matching?: boolean; // ignored for now
}
```

Still needs something to represent how these ConditionRows are releated to each other.

src/store/getters.ts
```diff
@@ -1,12 +1,20 @@
-import { RootState } from './RootState';
+import RootState from './RootState';
 import QueryRepresentation from '@/sparql/QueryRepresentation';
 import Property from '@/data-model/Property';
 
 export default {
-	query( { property, value }: RootState ): QueryRepresentation {
-		return { property, value };
+	query( rootState: RootState ): QueryRepresentation {
```
jakobw, line 6: Ah yes, this getter is already effectively one half of this two-way transformation between store state and QueryRepresentation. Cool!

src/store/RootState.ts
```diff
@@ -1,4 +1,13 @@
-import QueryRepresentation from '@/sparql/QueryRepresentation';
+export default interface RootState {
+	conditionRow: ConditionRow;
+}
 
-export interface RootState extends QueryRepresentation {
+export interface ConditionRow {
+	propertyData: {
```
jakobw, line 6: I was about to question the `Data` suffix on both `propertyData` and `valueData` but it makes sense as soon as the aforementioned meta data is also in here :+1:

jakobw (commented):
I like where this is going! The additions to `QueryRepresentation` and the `RootState` interface make sense to me. Thanks! :100:

jakobw:
> I think in the long run we could end up with something like [...]

Yup, looks reasonable. Probably with lists of lists for the distinction between "and"-groups and "or"-groups of conditions. And conditions probably become some sort of union type like

```ts
type Condition = PropertyValueCondition | PropertyAnyValueCondition | PropertyDataRangeCondition | PropertyCoordinatePolygonCondition | ...
```

jakobw (approved):

micgro42:
Todo for myself:

~~- [ ] add test for `query()` getter as that starts to become non-trivial. Then this should be good to be merged~~

Decided to merge this now and add the test in another PR.

micgro42:
Test added in #48

PR #45, opened 2020-11-05
Stricter and more reasonable linting

micgro42:
This should ideally be reviewed on a per-commit basis.

Biggest points:

- requires a trailing comma instead of forbidding it
- fails in CI with code that would cause warnings or has fixable errors
- don't apply liniting during `docker-compose up`/`npm run serve`


jakobw (approved):
LGTM :100:

micgro42:
Merging it as it has been approved.

@Ladsgroup @guergana @bereket-WMDE You may want to have a look at the changes in this PR in case that the configuration of your IDE needs adjustment.

PR #47, opened 2020-11-06
Upgrade @wmde/wikit-vue-components

micgro42:
This will enable us to use the new Lookup and Message components.

Also we no longer have to import from the src/ directory.

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -18,4 +18,5 @@ export default Vue.extend( {
 
 <style lang="scss">
 @import '~ress';
+@import '~@wmde/wikit-vue-components/dist/wikit-vue-components.css';
 </style>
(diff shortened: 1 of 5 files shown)


src/App.vue
```diff
@@ -18,4 +18,5 @@ export default Vue.extend( {
 
 <style lang="scss">
 @import '~ress';
+@import '~@wmde/wikit-vue-components/dist/wikit-vue-components.css';
```
Ladsgroup, line 21: This is so weird. How the styles even worked? I have a feeling we already import this somewhere but I can't find it :/

Ladsgroup: Maybe by importing the component, we automatically import its style as well (I'm new to Vue). If that's the case, what's the point of this import?

micgro42: Yes and No. Previously, we had imported from the `src/` directory and compiled them ourselves

```
- import TextInput from '@wmde/wikit-vue-components/src/components/TextInput.vue';
```

But now we are importing them, as we should, from `dist/`:

```
+ import { TextInput } from '@wmde/wikit-vue-components';
```

I agree that it is weird and feels wrong that we have to import the styles separately. I'm not sure that this needs doing for vuetify as well?

micgro42:
Rebased.

src/App.vue
Ladsgroup: It makes sense now, thanks for the explanation. I think it's good for now.

Ladsgroup (approved):

PR #52, opened 2020-11-09
A components test directory

micgro42:
As we are going to extract more sub-components, we should collect their tests in a dedicated directory, mirroring the structure in src/.


Ladsgroup (approved):
Thanks!

PR #53, opened 2020-11-09
Use Message component

Ladsgroup:
The component is not exposed yet, and some changes to scss values are
not in the release yet. All needs to be fixed later.

Bug: T266789


Ladsgroup:
This should be draft, given that the message component is not yet exposed.

guergana (approved):
Looks great!

PR #55, opened 2020-11-10
Add error reporting to each field

Ladsgroup:
Bug: T267576

### DIFF
--- a/src/components/QueryBuilder.vue
+++ b/src/components/QueryBuilder.vue
@@ -13,12 +13,14 @@
 					class="querybuilder__rule__property"
 					label="Property"
 					:value="selectedItem.label"
+					:error="fieldErrors.property"
 					placeholder="Enter a property" />
 				<TextInput
 					class="querybuilder__rule__value"
 					label="Value"
 					ref="value"
 					v-model="textInputValue"
+					:error="fieldErrors.value"
 					placeholder="Enter a value" />
 			</div>
 			<div class="querybuilder__run">
@@ -40,6 +42,7 @@ import QueryResult from '@/components/QueryResult.vue';
 import buildQuery from '@/sparql/buildQuery';
 import Property from '@/data-model/Property';
 import { mapState } from 'vuex';
+import Error from '@/data-model/Error';
 
 export default Vue.extend( {
 	name: 'QueryBuilder',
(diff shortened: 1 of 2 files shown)


tests/unit/components/QueryBuilder.spec.ts
```diff
@@ -52,4 +52,37 @@ describe( 'QueryBuilder.vue', () => {
...
+		expect( wrapper.findAllComponents( TextInput ).at( 0 ).props( 'error' ) ).toStrictEqual( {
+			type: 'error',
+			message: 'Property Error Message!',
+		} );
+		expect( wrapper.findAllComponents( TextInput ).at( 1 ).props( 'error' ) ).toStrictEqual( {
+			type: 'warning',
+			message: 'Value Warning Message!',
+		} );
+
+	} );
```
micgro42, line 87: I feel this could have tests that check for the correct errors being selected depending on the store state. OTOH, if we move this logic to an action soonish anyway, then maybe we can add it at that point.

src/components/QueryBuilder.vue
```diff
@@ -63,6 +70,18 @@ export default Vue.extend( {
 			}
 
 			if ( !this.selectedItem || !this.textInputValue ) {
+				if ( !this.selectedItem ) {
```
micgro42, line 73: I'm not particularly happy about adding more business logic to this component. But even if we add it to an action now, we still have to touch it again when we implement adding more conditions, so we might as well keep it here until then.

tests/unit/components/QueryBuilder.spec.ts
Ladsgroup: Yes, exactly. I will refactor the whole thing later.

src/components/QueryBuilder.vue
Ladsgroup: Indeed. My plan is to actually move everything into a js service like "Validator" or something which would accept a Form object and returns errors (I'm not sure about vue store action though but I can be convinced otherwise, as I'm not really expert in frontend) but this is all for future (and I will do it once the component becomes a bit stable, currently it's being changed a lot)

micgro42:
Cool, thank you :)

micgro42 (approved):

PR #57, opened 2020-11-11
Fix validator of property values

Ladsgroup:
Currently, if the property is empty, the property returns an empty
object that has "id" property and we can check against that instead.

Without this patch, it thinks the property is filled all the time.

Also, renaming the property to its correct name.


micgro42:
Nope, this is changing it at the wrong level. The state should not contain an empty id

Ladsgroup:
> Nope, this is changing it at the wrong level. The state should not contain an empty id

Done

micgro42 (approved):

PR #58, opened 2020-11-11
Move out validation logic out of QueryBuilder component

Ladsgroup:
This is business logic and should not exist there and the component is
already getting really big.

Also, introducing several interfaces to formalize relationship of
components and business domain.

### DIFF
--- a/src/components/QueryBuilder.vue
+++ b/src/components/QueryBuilder.vue
@@ -43,6 +43,7 @@ import SearchResult from '@/data-access/SearchResult';
 import Property from '@/data-model/Property';
 import Error from '@/data-model/Error';
 import buildQuery from '@/sparql/buildQuery';
+import Validator, { ValidationResult } from '@/form/Validator';
 
 export default Vue.extend( {
 	name: 'QueryBuilder',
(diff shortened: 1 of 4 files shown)


src/components/QueryBuilder.vue
```diff
@@ -57,42 +58,21 @@ export default Vue.extend( {
...
+			const formValues = {
+				property: this.selectedProperty,
+				value: this.textInputValue,
+			};
+			const validator = new Validator( formValues );
+			const validationResult = validator.validate();
+			this.errors = validationResult.formErrors;
+			this.$store.dispatch( 'setErrors', validationResult.formErrors );
+			this.fieldErrors = validationResult.fieldErrors;
+			return validationResult;
```
micgro42: It should not change the state _and_ return something. This violates command-query-separation. We can do without returning the result here and just check `this.errors` in the if-condition below.

micgro42 (commented):
I think it goes into the right direction, thank you!

I still have to look at the tests

src/components/QueryBuilder.vue
Ladsgroup: You're absolutely right. Done now.

tests/unit/form/Validator.spec.ts
```diff
@@ -0,0 +1,95 @@
+import FormValues from '@/form/FormValues';
+import Validator, { ValidationResult } from '@/form/Validator';
+
+describe( 'validator', () => {
+	it( 'full form', () => {
```
micgro42: This could be a bit more verbose. The standard form for these titles is something like
```suggestion
	it( 'returns no errors with a complete form', () => {
```

tests/unit/form/Validator.spec.ts
```diff
@@ -0,0 +1,95 @@
...
+			{
+				type: 'error',
+			},
+		];
+
+		const validator = new Validator( formValues );
+		const validationResult = validator.validate();
+
+		expect( validationResult.formErrors ).toHaveLength( 1 );
+		expect( validationResult.formErrors[ 0 ].type ).toBe( formErrors[ 0 ].type );
```
micgro42, line 43: I think it makes more sense to construct the actual error objects here, with the literal message for now, and to assert them with `.toStrictEqual`. Then those 5 assertions here can be replaced with two.

We can later replace the message with the message key when the i18n story is done.

micgro42 (changes requested):
Thank you for updating the PR! I have some small requests for the tests and then this is good to go.

tests/unit/form/Validator.spec.ts
Ladsgroup: That's exactly why I did five asserts instead of one. It's not just i18n itself, it's also repeating the error message over and over, meaning if we want to change the message (even a missing dot), we need to change lots of tests and I don't like such tight binding.

micgro42: Mh, I can accept that. But then I'll add a AC to the i18n ticket to fix these tests when we have the i18n implemented.

micgro42: => https://phabricator.wikimedia.org/T268014

Ladsgroup: Thanks!

micgro42 (approved):

PR #59, opened 2020-11-11
Make the "Run query" button translatable

Ladsgroup:
This lays the groundwork for i18n using vue-banana-i18n

Basically, having the i18n files exposed as public files and downloading
them when building the application and showing the results once it's
there.

Bug: T267730


micgro42:
Looks like a good approach in principle. It needs some more polishing (using async/await, missing newlines at the end of file, etc.).

I think the biggest draw-back is the missing fallback for incomplete translations, but we can implement that when it is requested, which is likely to be soon.

Also, but probably at a later date, this will need some extra config for when we are not running at the root of a domain but at `www.example.org/query-builder/...`

src/App.vue
```diff
@@ -1,14 +1,33 @@
...
 
 <script lang="ts">
 import Vue from 'vue';
 import QueryBuilder from '@/components/QueryBuilder.vue';
+import i18n from 'vue-banana-i18n';
 
 export default Vue.extend( {
+	data() {
+		return {
+			isi18nLoaded: false as boolean,
```
bereket-WMDE, line 15: isn't just 'false' enough here ?

src/App.vue
```diff
@@ -1,14 +1,33 @@
 <template>
 	<div id="app">
-		<QueryBuilder/>
+		<QueryBuilder v-if="isi18nLoaded"/>
```
bereket-WMDE, line 3: if this is false, does that mean nothing is loaded ? 
as micheal suggested, we need some sort of Fallback mechanism.

public/i18n/en.json
```diff
@@ -0,0 +1,3 @@
+{
+  "query-builder-run-query": "Run query"
+}
```
bereket-WMDE: as the documentation ( https://cli.vuejs.org/guide/html-and-static-assets.html#url-transform-rules ) suggests, using public folder directly, might not be the best approach. 
what are the arguments we have for not making it part of the module ?

src/App.vue
Ladsgroup: It seems not, TypeScript was complaining.

src/App.vue
Ladsgroup: Yes but the i18n should be loaded for that language first, we can implement falling back chain of languages later (you load that language and English and if the message doesn't exist, vue-banana automatically falls back to English)

public/i18n/en.json
Ladsgroup: I understand but OTOH, if we want to add it to webpack, we have to add all languages to it and it would make it explode, I highly recommand keeping this part as it is (given the special way i18n works in Wikimedia).

bereket-WMDE (approved):
In the spirit of moving things along, I'd approve this PR. 
just fix up the missing "no new line at the end of file" warnings.
we should keep @micgro42 suggestions in mind, though.

public/i18n/en.json
micgro42, line 3: Missing new-line at the end of file

public/i18n/qqq.json
```diff
@@ -0,0 +1,3 @@
+{
+  "query-builder-run-query": "Text on the button that runs the query"
+}
```
micgro42: Missing new-line at the end of file

public/i18n/en.json
micgro42: Missing new-line at the end of file

src/App.vue
```diff
@@ -1,14 +1,33 @@
...
 
 export default Vue.extend( {
+	data() {
+		return {
+			isi18nLoaded: false as boolean,
+		};
+	},
+	beforeCreate(): void {
+		fetch( 'i18n/en.json' )
+			.then( function ( response ) {
```
micgro42: This method should use `async`/`await` instead.

src/App.vue
```diff
@@ -1,14 +1,33 @@
...
+				return response.json();
+			} ).then( ( messages ) => {
+				Vue.use( i18n, {
+					locale: 'en',
+					messages,
+				} );
+				// TODO: Fix the typing warning
+				( this as any ).isi18nLoaded = true;
```
micgro42: This typing problem might already go away when we use async/await :crossed_fingers:

public/i18n/en.json
Ladsgroup: Done

public/i18n/qqq.json
Ladsgroup: Done

src/App.vue
Ladsgroup: Didn't :(

src/App.vue
Ladsgroup: Done

micgro42 (approved):
Thanks! We'll figure they types out later

Ladsgroup:
> Thanks! We'll figure they types out later

This seems to happen a lot (google says so, one way to fix it is that you need set the return type of all methods but we did that already). Here's an example: https://github.com/vuejs/vetur/issues/2131

PR #60, opened 2020-11-12
Accept all type of triples not truthy ones only

Ladsgroup:
The predicate should change from "wdt" to "p/ps" (SequencePath[1])

[1]: https://www.w3.org/TR/sparql11-query/#propertypaths

Bug: T266240


Ladsgroup:
If you want to test this. Try DMOZ ID and "World/Svenska/Kultur/Litteratur/Genre/Science_fiction_och_fantasy/Författare/Adams,_Douglas/" as value (without quotes)

micgro42 (approved):

PR #61, opened 2020-11-12
[WIP] Added ie11 to the list of supported browsers

guergana:
We have to explicitly define this because Internet Explorer does not have modern JS features.

More info here: https://cli.vuejs.org/guide/browser-compatibility.html#browserslist
and here: https://github.com/browserslist/browserslist

### DIFF
--- a/.browserslistrc
+++ b/.browserslistrc
@@ -1,3 +1,7 @@
+#Browsers that we support
+
 > 1%
 last 2 versions
 not dead
+
+ie 11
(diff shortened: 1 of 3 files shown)


.browserslistrc
```diff
@@ -1,3 +1,7 @@
+#Browsers that we support
+
 > 1%
 last 2 versions
 not dead
+
+ie 11
```
micgro42: This could have a trailing newline

micgro42 (approved):

guergana:
only adding `IE` to the `.browserslistrc` file is not working. This most likely needs to be made in combination with other commands but the documentation is not very clear about this. To be continued...

micgro42:
It could very well be that we have to add some polyfills manually, such as for `fetch`

guergana:
Closing because we've dropped IE11 support for now.

PR #65, opened 2020-11-18
Migrate user-facing texts to i18n

Ladsgroup:
Bug: T268014


public/i18n/qqq.json
```diff
@@ -1,3 +1,16 @@
 {
-  "query-builder-run-query": "Text on the button that runs the query"
+  "query-builder-heading": "Title of the page",
+  "query-builder-description": "A short description on what query builder is",
+  "query-builder-find-all-items": "header for part of the form finding all items",
+  "query-builder-run-query": "Text on the button that runs the query",
+  "query-builder-input-value": "Text of value of the property in the form",
+  "query-builder-input-enter-value": "Placeholder of the value input",
+  "query-builder-property-lookup-enter-property": "Label of the property input",
+  "query-builder-property-lookup-property": "Placeholder of the property input",
```
micgro42: The `query-builder-property-lookup-property` and `query-builder-property-lookup-enter-property` are mixed up. It would probably be better if the message names could explicitly include their purpose. So calling them something like `query-builder-property-lookup-label` and `query-builder-property-lookup-placeholder`

public/i18n/qqq.json
```diff
@@ -1,3 +1,16 @@
 {
-  "query-builder-run-query": "Text on the button that runs the query"
+  "query-builder-heading": "Title of the page",
+  "query-builder-description": "A short description on what query builder is",
+  "query-builder-find-all-items": "header for part of the form finding all items",
+  "query-builder-run-query": "Text on the button that runs the query",
+  "query-builder-input-value": "Text of value of the property in the form",
```
micgro42: This name should be `query-builder-input-value-label` or maybe `query-builder-value-input-label`. Also the explanation could be a bit clearer. What about "Label of the value input" or "Label of the property value input"?

public/i18n/en.json
```diff
@@ -1,3 +1,16 @@
 {
-  "query-builder-run-query": "Run query"
+  "query-builder-heading": "Simple Query Builder",
+  "query-builder-description": "Welcome to the test system of the Simple Query Builder. Please note that this is a work in progress, not all features are included, and it can sometimes be broken. [$1 Feedback is welcome here.]",
```
micgro42: Are we sure we want _this_ part translated? I think it is fine if that specific copy stays in the query builder for now, at least until we have a decision on https://phabricator.wikimedia.org/T265240

How do you see that?

micgro42 (changes requested):
Nice work. I like the much clearer tests :).

public/i18n/qqq.json
Ladsgroup: You're right. Done now.

public/i18n/qqq.json
Ladsgroup: You're right. I changed it. I changed it to `query-builder-input-value-label` as more I wanted it to have tree like structure. "label of value of input part of query builder"

public/i18n/en.json
Ladsgroup: You are right, it would put too much burden and translators on something we are going to throw away soon. It's just weird to see English text at middle of an Arabic interface for example but you're right, we'll migrate it later.

micgro42 (approved):

public/i18n/en.json
micgro42: Yeah, I sure believe that this is looks weird. Let's worry about it again when we actually have a way to switch the language

public/i18n/qqq.json
micgro42: Makes sense. Thanks!

public/i18n/qqq.json
micgro42: :+1:

PR #66, opened 2020-11-19
Create ValueTypeDropdown component in QB

bereket-WMDE:
Bug: T267735

### DIFF
--- a/src/components/QueryBuilder.vue
+++ b/src/components/QueryBuilder.vue
@@ -13,6 +13,9 @@
 					v-model="selectedProperty"
 					:error="fieldErrors.property"
 				/>
+				<ValueTypeDropDown
+					@input="onInput"
+				/>
 				<TextInput
 					class="querybuilder__rule__value"
 					label="Value"
@@ -41,6 +44,7 @@ import { mapState } from 'vuex';
 import { Button, TextInput } from '@wmde/wikit-vue-components';
 
 import PropertyLookup from '@/components/PropertyLookup.vue';
+import ValueTypeDropDown from '@/components/ValueTypeDropDown.vue';
 import QueryResult from '@/components/QueryResult.vue';
 import SearchResult from '@/data-access/SearchResult';
 import Property from '@/data-model/Property';
@@ -58,9 +62,13 @@ export default Vue.extend( {
 				property: null as null | Error,
 				value: null as null | Error,
 			},
+			selectedOption: '',
 		};
 	},
 	methods: {
+		onInput( value: string ): void {
+			this.selectedOption = value;
+		},
 		validate(): void {
 			const formValues = {
 				property: this.selectedProperty,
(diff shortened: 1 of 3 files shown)


src/components/QueryBuilder.vue
```diff
@@ -58,9 +63,18 @@ export default Vue.extend( {
 				property: null as null | Error,
 				value: null as null | Error,
 			},
+			optionItems: [
```
micgro42: Those can be inside the ValueTypeDropdown as well. They Idea of this component is to encapsulate all the things that are specific to this particular Dropdown.

micgro42 (commented):
Looks good so far!

src/components/QueryBuilder.vue
```diff
@@ -58,9 +63,18 @@ export default Vue.extend( {
...
+				{ value: 'without' },
+				{ value: 'with' },
+				{ value: 'regardless' },
```
guergana: It looks good. We probably need to check if it is with these options or "matching" and "regardless of value" as indicated in this ticket. https://phabricator.wikimedia.org/T265245

src/components/QueryBuilder.vue
bereket-WMDE: right. I moved it out of the query builder

src/components/QueryBuilder.vue
bereket-WMDE: got an updated from charlie and adjusted it accordingly

src/components/ValueTypeDropDown.vue
```diff
@@ -0,0 +1,51 @@
...
+<script lang="ts">
+import Vue from 'vue';
+
+export default Vue.extend( {
+	name: 'ValueTypeDropDown',
+	data() {
+		return {
+			selected: null,
+			optionItems: [
+				{ value: 'matching' },
```
micgro42, line 24: When we switch to using the Dropdown, we probably will want to replace those with constants from an enum. But it is fine for now

src/components/QueryBuilder.vue
```diff
@@ -13,6 +13,9 @@
 					v-model="selectedProperty"
 					:error="fieldErrors.property"
 				/>
+				<ValueTypeDropDown
+					@input="onInput"
```
micgro42, line 17: We probably want this to be a `v-model` as well at some point, so that it can be _set_ when filling the QueryBuilder from an URL. But than could also just be done as part of https://phabricator.wikimedia.org/T267736

micgro42 (approved):

src/components/ValueTypeDropDown.vue
```diff
@@ -0,0 +1,51 @@
...
+				{ value: 'regardless of value' },
+				{ value: 'without' },
```
guergana, line 26: nice! updated on my pr as well

src/components/QueryBuilder.vue
bereket-WMDE: yes, i Implemented that on T267736

PR #67, opened 2020-11-19
Adjust QueryRepresentation and SPARQL query of any / regardless state

guergana:
+ added enum type PropertyValueRelation
+ updated condition instances to use PropertyValueRelation
+ updated QueryObjectBuilder by adding "any value" case
+ added test to confirm generation of any value case query

### DIFF
--- a/src/data-model/PropertyValueRelation.ts
+++ b/src/data-model/PropertyValueRelation.ts
@@ -1,0 +1,6 @@
+enum PropertyValueRelation {
+	Matching = 'matching',
+	Regardless = 'regardless of value',
+}
+
+export default PropertyValueRelation;
--- a/src/sparql/QueryObjectBuilder.ts
+++ b/src/sparql/QueryObjectBuilder.ts
@@ -1,6 +1,7 @@
 import rdfNamespaces from '@/sparql/rdfNamespaces';
 import QueryRepresentation from '@/sparql/QueryRepresentation';
-import { SelectQuery } from 'sparqljs';
+import { SelectQuery, Term } from 'sparqljs';
+import PropertyValueRelation from '@/data-model/PropertyValueRelation';
 
 export default class QueryObjectBuilder {
 	private queryObject: SelectQuery;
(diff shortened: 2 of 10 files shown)


src/sparql/QueryObjectBuilder.ts
```diff
@@ -52,6 +54,46 @@ export default class QueryObjectBuilder {
 			},
 		];
 
+		if ( queryRepresentation.condition.propertyValueRelation === PropertyValueRelation.NotMatching ) {
```
Ladsgroup: This is repeating a lot of query building logic from the above. Can't it have some cleaner way to do it?

guergana: Hi, I am aware of this, will do in next iteration.

tests/unit/sparql/buildQuery.spec.ts
```diff
@@ -1,14 +1,51 @@
...
+		} } ) ).toEqual( expectedQuery );
+	} );
+
+	it( 'builds a query from a property with "any value" selected', () => {
+		const propertyId = 'P666';
+		const propertyValueRelation = PropertyValueRelation.Regardless;
+
+		expect( buildQuery( { condition: {
+			propertyId,
+			value: '',
```
guergana, line 23: I first tried not sending the value but it got messy with the types from the Terms tripleObject in `src/sparql/QueryObjectBuilder.ts `. Happy to read suggestions.

micgro42: Mh, I think it is okay to use it like that for now. Maybe we could also use `null`, but an empty string is also just fine here as that should not be possible by user input.

src/data-model/PropertyValueRelation.ts
```diff
@@ -0,0 +1,6 @@
+enum PropertyValueRelation {
```
micgro42, line 1: I like it!

micgro42 (approved):

PR #68, opened 2020-11-19
Add ability to change language with uselang=foo

Ladsgroup:
- Fixing type error by changing the hook to created() so the data would
   be available
 - Handing cases where the response is non-existent or not parsable
 - Requesting English regardless since it needs to fall back if a
   certain message doesn't exist in that language
 - Using uselang= to make it consistent with rest of Wikimedia
   applications

Bug: T267732

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -1,5 +1,5 @@
 <template>
-	<div id="app">
+	<div id="app" :lang="lang" dir="auto">
 		<QueryBuilder v-if="isi18nLoaded"/>
 	</div>
 </template>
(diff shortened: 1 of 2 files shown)


src/App.vue
```diff
@@ -13,21 +13,44 @@ export default Vue.extend( {
...
+				this.lang = urlParams.get( 'uselang' ) as string;
+			}
+
+			const responseEn = await fetch( 'i18n/en.json' );
+			const messages: { [key: string]: { [key: string]: string} } = {
+				en: await responseEn.json(),
 			};
+
+			if ( this.lang !== 'en' ) {
+				const responseLang = await fetch( 'i18n/' + this.lang + '.json' ).catch(
```
micgro42: `await` + `catch` is an unusual combination. Why not use try-catch here as well?

src/App.vue
```diff
@@ -13,21 +13,44 @@ export default Vue.extend( {
...
+					() => {
+						// TODO: Show a user-facing notification instead
+						console.warn( 'The language requested could not be retrieved, falling back to English' );
+						messages[ this.lang ] = {};
+					},
+				) as Response;
+				if ( responseLang.ok ) {
```
micgro42: Are we sure that `responseLang` is defined if the catch above is triggered?

micgro42 (commented):
I think I want most of this code to move into a dedicated repository and maybe some helper function for parsing the search params. But that could be done in a follow-up.

Also, is the current solution already using en as a fallback, or do we somehow define that somewhere explicitly?

Ladsgroup:
> I think I want most of this code to move into a dedicated repository and maybe some helper function for parsing the search params. But that could be done in a follow-up.
> 

Yeah, the funny is that it started as a tiny thing and slowly grew and grew. I will migrate it date-access in a follow up
> Also, is the current solution already using en as a fallback, or do we somehow define that somewhere explicitly?

I'm not sure what you mean. English has been always the fallback everywhere.

src/App.vue
Ladsgroup: yeah, I'll do it

src/App.vue
Ladsgroup: yeah it needs clean up

micgro42:
> > Also, is the current solution already using en as a fallback, or do we somehow define that somewhere explicitly?
> 
> I'm not sure what you mean. English has been always the fallback everywhere.

Ok, so it is hard-coded (or at least the default) in vue-banana-i18n to look for `en`-messages if the message key doesn't exist in the selected/preferred language?

Ladsgroup:
> > > Also, is the current solution already using en as a fallback, or do we somehow define that somewhere explicitly?
> > 
> > 
> > I'm not sure what you mean. English has been always the fallback everywhere.
> 
> Ok, so it is hard-coded (or at least the default) in vue-banana-i18n to look for `en`-messages if the message key doesn't exist in the selected/preferred language?

It's not in vue-banana-i18n, it's in banana-i18n and it has a list of fallback chains (https://github.com/wikimedia/banana-i18n/blob/master/src/languages/fallbacks.json so for example lots of Iranian languages fall back to Persian before falling back to English) and there English is default fallback of all languages/

Ladsgroup:
https://github.com/wikimedia/banana-i18n/blob/master/src/index.js#L14

micgro42 (approved):

PR #69, opened 2020-11-20
Make it possible to add the value type dropdown value to the state

bereket-WMDE:
will be merged after [this](https://github.com/wmde/query-builder/pull/66l) is merged to master

### DIFF
--- a/src/components/QueryBuilder.vue
+++ b/src/components/QueryBuilder.vue
@@ -16,7 +16,7 @@
 					:error="fieldErrors.property"
 				/>
 				<ValueTypeDropDown
-					@input="onInput"
+					v-model="selectedPropertyValueRelation"
 				/>
 				<TextInput
 					class="querybuilder__rule__value"
@@ -51,6 +51,7 @@ import ValueTypeDropDown from '@/components/ValueTypeDropDown.vue';
 import QueryResult from '@/components/QueryResult.vue';
 import SearchResult from '@/data-access/SearchResult';
 import Property from '@/data-model/Property';
+import PropertyValueRelation from '@/data-model/PropertyValueRelation';
 import Error from '@/data-model/Error';
 import buildQuery from '@/sparql/buildQuery';
 import Validator from '@/form/Validator';
@@ -65,17 +66,14 @@ export default Vue.extend( {
 				property: null as null | Error,
 				value: null as null | Error,
 			},
-			selectedOption: '',
 		};
 	},
 	methods: {
-		onInput( value: string ): void {
-			this.selectedOption = value;
-		},
 		validate(): void {
 			const formValues = {
 				property: this.selectedProperty,
 				value: this.textInputValue,
+				relation: this.selectedPropertyValueRelation,
 			};
 			const validator = new Validator( formValues );
 			const validationResult = validator.validate();
@@ -101,6 +99,14 @@ export default Vue.extend( {
 			},
 			set( selectedProperty: SearchResult ): void {
 				this.$store.dispatch( 'updateProperty', selectedProperty );
+			},
+		},
+		selectedPropertyValueRelation: {
+			get(): PropertyValueRelation {
+				return this.$store.getters.propertyValueRelation;
+			},
+			set( selectedPropertyValueRelation: string ): void {
+				this.$store.dispatch( 'updatePropertyValueRelation', selectedPropertyValueRelation );
 			},
 		},
 		textInputValue: {
--- a/src/store/RootState.ts
+++ b/src/store/RootState.ts
@@ -14,7 +14,7 @@ export interface ConditionRow {
 	valueData: {
 		value: string;
 	};
-	propertyValueRelation: {
+	propertyValueRelationData: {
 		value: PropertyValueRelation;
 	};
 }
--- a/src/store/index.ts
+++ b/src/store/index.ts
@@ -22,7 +22,7 @@ export function createStore( services: QueryBuilderServices ): Store<RootState> 
 				valueData: {
 					value: '',
 				},
-				propertyValueRelation: {
+				propertyValueRelationData: {
 					value: PropertyValueRelation.Matching,
 				},
 			},
--- a/tests/unit/store/getters.spec.ts
+++ b/tests/unit/store/getters.spec.ts
@@ -10,7 +10,7 @@ describe( 'getters', () => {
 				conditionRow: {
 					valueData: { value: 'foo' },
 					propertyData: { id: 'P123', label: 'abc' },
-					propertyValueRelation: { value: PropertyValueRelation.Matching },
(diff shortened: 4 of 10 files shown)


src/store/RootState.ts
```diff
@@ -13,4 +13,8 @@ export interface ConditionRow {
 	valueData: {
 		value: string;
 	};
+	// TODO: determine what the name should be
+	propertyValueRelationData: {
```
guergana, line 17: I named it propertyValueRelation :D now I see all of them have *data as part of the name

tests/unit/store/mutations.spec.ts
```diff
@@ -9,6 +9,7 @@ describe( 'mutations', () => {
 			conditionRow: {
 				valueData: { value: 'foo' },
 				propertyData: { id: 'P123', label: 'abc' },
+				propertyValueRelationData: { value: 'with' },
```
guergana: I think the value should be = 'matching' for clarity or enum `propertyValueRelation: {
					value: PropertyValueRelation.Matching,
				},`

tests/unit/store/mutations.spec.ts
```diff
@@ -24,6 +25,7 @@ describe( 'mutations', () => {
...
```
guergana: same here: value = 'matching' or with enum `propertyValueRelation: {
					value: PropertyValueRelation.Matching,
				},`

tests/unit/store/getters.spec.ts
```diff
@@ -9,6 +9,7 @@ describe( 'getters', () => {
 				conditionRow: {
 					valueData: { value: 'foo' },
 					propertyData: { id: 'P123', label: 'abc' },
+					propertyValueRelationData: { value: 'with' },
```
guergana: same value = 'matching'

guergana: once we make an enum for this we might as well do something like: 
`propertyValueRelation: PropertyValueRelation.Matching,
`

src/store/index.ts
```diff
@@ -21,6 +21,9 @@ export function createStore( services: QueryBuilderServices ): Store<RootState>
 				valueData: {
 					value: '',
 				},
+				propertyValueRelationData: {
+					value: '',
```
guergana: same here when we are using the enum:

`propertyValueRelation: {
					value: PropertyValueRelation.Matching,
				},`

guergana (changes requested):
looks very good. just a few comments on naming

src/store/RootState.ts
```diff
@@ -13,4 +13,8 @@ export interface ConditionRow {
...
+		value: string;
```
micgro42: As soon as #67 is merged, we could use that enum for this instead of string literals

src/components/QueryBuilder.vue
```diff
@@ -16,6 +16,7 @@
 					:error="fieldErrors.property"
 				/>
 				<ValueTypeDropDown
+					v-model="selectedPropertyValueRelation"
 					@input="onInput"
```
micgro42: I think we no longer need this `@input="onInput"`, right?

guergana:
If you rebase this now on master we can reuse the enum PropertyValueRelation and the names. :)

src/store/RootState.ts
guergana: Yes, agree

src/store/RootState.ts
bereket-WMDE: yea. will keep the *data and adjust your changes

src/components/QueryBuilder.vue
```diff
@@ -103,6 +100,14 @@ export default Vue.extend( {
 				this.$store.dispatch( 'updateProperty', selectedProperty );
 			},
 		},
+		selectedPropertyValueRelation: {
+			get(): string | null {
```
guergana: why is the return value `string | null`? isn't it always a string? I was assuming the default is = 'matching' instead of null. what do you think @micgro42 

after taking a second look i think the type should be `PropertyValueRelation`

micgro42: > after taking a second look i think the type should be PropertyValueRelation

I agree

micgro42:
The options defined in `ValueTypeDropDown` could now also be the ones from `PropertyValueRelation`.

Also, what I unfortunately didn't notice before, the `ValueTypeDropDown` needs to emit the selected values. The on-click handler isn't actually working in this case (please check for yourself!).

Since this is a native `<select>`, we need to do a little dance here: We can add a `v-model` to the `<select>` tag and maybe point it to that unused `selected` key in `data()`. Then we can update that by watching it and the value prop with something like:

```ts
	watch: {
		selected(): void {
			this.$emit( 'input', this.selected );
		},
		value( selectedOption: PropertyValueRelation ): void {
			this.selected = selectedOption;
		},
	}
```

This should be a bit cleaner when we actually have the Dropdown.

tests/unit/store/mutations.spec.ts
bereket-WMDE: updated

tests/unit/store/mutations.spec.ts
bereket-WMDE: updated

tests/unit/store/getters.spec.ts
bereket-WMDE: updated

src/store/index.ts
bereket-WMDE: updated

src/components/QueryBuilder.vue
bereket-WMDE: good catch. i removed it

src/components/QueryBuilder.vue
bereket-WMDE: @guergana good catch :+1:

bereket-WMDE:
> The options defined in `ValueTypeDropDown` could now also be the ones from `PropertyValueRelation`.
> 
> Also, what I unfortunately didn't notice before, the `ValueTypeDropDown` needs to emit the selected values. The on-click handler isn't actually working in this case (please check for yourself!).
> 
> Since this is a native `<select>`, we need to do a little dance here: We can add a `v-model` to the `<select>` tag and maybe point it to that unused `selected` key in `data()`. Then we can update that by watching it and the value prop with something like:
> 
> ```ts
> 	watch: {
> 		selected(): void {
> 			this.$emit( 'input', this.selected );
> 		},
> 		value( selectedOption: PropertyValueRelation ): void {
> 			this.selected = selectedOption;
> 		},
> 	}
> ```
> 
> This should be a bit cleaner when we actually have the Dropdown.



made some changes according to the suggestions

micgro42 (approved):

guergana (approved):
looks good.

PR #70, opened 2020-11-20
Adjust Query Representation and SPARQL Query for without state

guergana:
Accidentally worked on this before the ticket was pulled into the sprint. Will leave here for next week

+ added without state to enum
+ modified QueryRepresentation by adding without state
+ added test to verify that query is being generated correctly

### DIFF
--- a/src/sparql/QueryObjectBuilder.ts
+++ b/src/sparql/QueryObjectBuilder.ts
@@ -1,6 +1,6 @@
 import rdfNamespaces from '@/sparql/rdfNamespaces';
 import QueryRepresentation from '@/sparql/QueryRepresentation';
-import { SelectQuery, Term } from 'sparqljs';
+import { Pattern, SelectQuery, Term } from 'sparqljs';
 import PropertyValueRelation from '@/data-model/PropertyValueRelation';
 
 export default class QueryObjectBuilder {
@@ -32,6 +32,12 @@ export default class QueryObjectBuilder {
 				tripleObject = {
 					termType: 'Literal',
 					value: queryRepresentation.condition.value,
+				};
+				break;
+			case ( PropertyValueRelation.NotMatching ):
+				tripleObject = {
+					termType: 'Variable',
+					value: 'instance',
 				};
 				break;
 			case ( PropertyValueRelation.Regardless ):
@@ -72,6 +78,28 @@ export default class QueryObjectBuilder {
 				],
 			},
 		];
+
+		if ( queryRepresentation.condition.propertyValueRelation === PropertyValueRelation.NotMatching ) {
+			const filterCondition = {
+				type: 'filter',
+				expression: {
+					type: 'operation',
+					operator: '!=',
+					args: [
+						{
+							termType: 'Variable',
+							value: 'instance',
+						},
+						{
+							termType: 'Literal',
+							value: queryRepresentation.condition.value,
+						},
+					],
+				},
+			};
+
+			this.queryObject.where.push( filterCondition as Pattern );
+		}
 
 		return this.queryObject;
 	}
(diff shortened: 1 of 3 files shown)


guergana:
@Ladsgroup here, ready for review.

src/sparql/QueryObjectBuilder.ts
```diff
@@ -73,6 +79,28 @@ export default class QueryObjectBuilder {
 			},
 		];
 
+		if ( queryRepresentation.condition.propertyValueRelation === PropertyValueRelation.NotMatching ) {
```
micgro42, line 82: This method is getting quite long. But I think it is still ok and we can refactor it when we get to the multiple conditions.

tests/unit/sparql/buildQuery.spec.ts
```diff
@@ -14,6 +14,29 @@ describe( 'buildQuery', () => {
 		} } ) ).toEqual( `SELECT ?item WHERE { ?item (p:${propertyId}/ps:${propertyId}) "${value}". }` );
 	} );
 
+	it( 'builds a query from a property and a string value with not matching selected', () => {
+		const propertyId = 'P666';
+		const value = 'blah';
+		const propertyValueRelation = PropertyValueRelation.NotMatching;
+
+		/*
+		* adding an array here to circumvent backticks + indentation problems between
```
micgro42: I think the more scalable solution might be to just use backticks here and not worry about spaces.

In the assert, we can then replace any whitespace `\s+` with a single space `' '` for _both_ the expected and the actual query. See [String.prototype.replaceAll()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replaceAll)

guergana: the problem is that if you use back ticks then the normal indentation of the code gets added as tabs and it breaks the test. This will probably illustrate the problem better: https://erikmartinjordan.com/avoid-tabs-multiline-string-backticks
Other suggested options would be to use a library, but I thought that might be better solved using native js. https://stackoverflow.com/questions/25924057/multiline-strings-that-dont-break-indentation

guergana: Alternatively we can not indent the code but I don't think that's an ideal solution and it would create lint errors. Also leaving it all in one line and adding `\n` makes the line too long. I tried all different ways and this seems the most decent solution.

Ladsgroup: Given that these are tests, I wouldn't worry too much. If it gets really big, you can move it a file and read that instead IMO

Ladsgroup:
If Michael is happy, I think it's fine to merge this (my virtual approval)

tests/unit/sparql/buildQuery.spec.ts
micgro42: As I wrote above, we can replace all the whitespace, including any tabs and newlines and such, with a single space before actually asserting the equality. That should ensure that the test still works and still tests that the actual code is equal, right?

guergana: I tried with trim and it didn't work well, will try with replaceAll

guergana: for the record, `replaceAll` is not supported by all browsers and causes the tests to fail. did the same with `replace`. https://caniuse.com/?search=replaceAll

micgro42: I'm curious to see how replaceAll failed, but I'm fine with just using replace.

Now all that still needs doing is removing this now obsolete commend and then the PR can be approved and merged :)

guergana: It says `replaceAll` is not a function as the reason for the test to fail. You can try it out. :) 
Will delete the comment, thanks for the heads-up.

micgro42 (approved):

PR #71, opened 2020-11-20
Disable Input field depending on state of ValueTypeDropdown

guergana:
- updated validator to not throw error on missing value when regarless is chosen
- wired valuetextinput to ValueTypeDropdown
- updated Validator to include propertyValueRelation
- added test for form validator to check for missing value in case of regardless
and matching

it is based on [PR 69](https://github.com/wmde/query-builder/pull/69). Will rebase once it is finished and merged.

### DIFF
--- a/src/components/QueryBuilder.vue
+++ b/src/components/QueryBuilder.vue
@@ -25,7 +25,9 @@
 					v-model="textInputValue"
 					:error="fieldErrors.value ?
 						{message: $i18n(fieldErrors.value.message), type: fieldErrors.value.type}: null"
-					:placeholder="$i18n('query-builder-input-value-placeholder')" />
+					:placeholder="$i18n('query-builder-input-value-placeholder')"
+					:disabled="selectedPropertyValueRelation === propertyValueRelation.Regardless"
+				/>
 			</div>
 			<div class="querybuilder__run">
 				<Button
@@ -66,6 +68,8 @@ export default Vue.extend( {
 				property: null as null | Error,
 				value: null as null | Error,
 			},
+			selectedOption: '',
+			propertyValueRelation: PropertyValueRelation,
 		};
 	},
 	methods: {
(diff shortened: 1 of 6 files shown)


src/form/Validator.ts
```diff
@@ -29,7 +30,9 @@ export default class Validator {
 			return validationResult;
 		}
 
-		if ( !this.formValues.property || !this.formValues.value ) {
+		if ( !this.formValues.property ||
+			( !( this.formValues.propertyValueRelation === PropertyValueRelation.Regardless ) &&
```
Ladsgroup: I feel this condition here is getting quite complex and hard to read. Is it possible to split it? maybe move some it into a new method. etc.

Ladsgroup (commented):
One nitpick and it's good to go.

src/form/Validator.ts
guergana: I agree. I did struggle a bit with this. Do you have any suggestions on how to restructure this conditional to make it better? I tried to keep it similar to what was already there but yes it is very unreadable now.

Ladsgroup: You can easily take out the property part (and define an incomplete form boolean and use that later to add the form error). I can add a follow up or something similar in a follow up PR.

micgro42: ~~Also, there should be a test for this change to the Validator~~ That is there, somehow I missed it when looking through the files after getting distracted.

src/components/QueryBuilder.vue
```diff
@@ -25,7 +25,8 @@
 					v-model="textInputValue"
 					:error="fieldErrors.value ?
 						{message: $i18n(fieldErrors.value.message), type: fieldErrors.value.type}: null"
-					:placeholder="$i18n('query-builder-input-value-placeholder')" />
+					:placeholder="$i18n('query-builder-input-value-placeholder')"
+					:disabled="selectedPropertyValueRelation === propertyValueRelation.Regardless" />
```
micgro42: We probably should put the `/>` on its own line. Then we don't have those extra line changes when we have to add another attribute.

micgro42 (commented):
Either in this or in the previous commit we have to also set the content of the value input to the empty string.

Probably better done in this commit as it already touches that element.

src/components/QueryBuilder.vue
guergana: Good idea. Changed.

Ladsgroup (approved):

PR #73, opened 2020-11-23
Add banana tests

Ladsgroup:
- Make sure all messages start with "query-builder-" prefix to avoid
   clashes in translatewiki.
 - Make sure all en.json messages have description in qqq.json and we
   don't have any orphan descriptions either.

Bug: T263553

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4596,9 +4596,9 @@
       }
     },
     "@types/node": {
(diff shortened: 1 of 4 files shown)


guergana (approved):
Still scared about using `fs` but if it's the only way to go. :+1:

PR #74, opened 2020-11-24
Add a hidden label for screen readers

Ladsgroup:
This fixes the a11y issue, also given that vue2 doesn't support multiple
elements in one component, wrapping everything under a nice <div>

Bug: T265245

### DIFF
--- a/src/components/ValueTypeDropDown.vue
+++ b/src/components/ValueTypeDropDown.vue
@@ -1,17 +1,21 @@
 <template>
-	<select
-		class="querybuilder__dropdown"
-		v-model="selected"
-	>
-		<option
-			class="querybuilder__dropdown-option"
-			v-for="(optionItem, index) in optionItems"
-			:value="optionItem"
-			:key="index"
+	<div class="querybuilder__dropdown">
+		<label class="querybuilder__dropdown-label">Value Type</label>
+		<select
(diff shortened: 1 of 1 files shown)


src/components/ValueTypeDropDown.vue
```diff
@@ -1,17 +1,21 @@
...
-		class="querybuilder__dropdown"
-		v-model="selected"
-	>
-		<option
-			class="querybuilder__dropdown-option"
-			v-for="(optionItem, index) in optionItems"
-			:value="optionItem"
-			:key="index"
+	<div class="querybuilder__dropdown">
+		<label class="querybuilder__dropdown-label">Value Type</label>
```
micgro42: I don't think that this will satisfy the test, it probably needs a `for="someId"` attribute to actually link it to the select. _Alternatively_, the select could maybe get an [`aria-labeledby` attribute]9https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-labelledby_attribute)

Ladsgroup: of course, you're right. I will work on it.

Ladsgroup: Done now.

micgro42 (approved):

PR #75, opened 2020-11-24
Add limited support messages

Ladsgroup:
Bug: T268446


guergana (approved):
Looks good. nice text!

PR #76, opened 2020-11-24
Use specialized method to select an <option> in test

micgro42:
~~The setSelected() method was added in vuejs/vue-test-utils#557, but apparently never documented. It seems stable and intended for end-user use nonetheless, so it is probably just an oversight documentation wise.~~ The `setSelected()` is actually documented in https://vue-test-utils.vuejs.org/api/wrapper/#setselected -- no idea why I didn't find it when I first looked.

This PR should make the test last touched in #72 more robust

### DIFF
--- a/tests/unit/components/ValueTypeDropDown.spec.ts
+++ b/tests/unit/components/ValueTypeDropDown.spec.ts
@@ -8,15 +8,10 @@ describe( 'ValueTypeDropDown.vue', () => {
 		const optionItems = PropertyValueRelation;
 		const wrapper = shallowMount( ValueTypeDropDown );
 
-		wrapper.setData( {
-			selected: optionItems.Matching,
-			optionItems: optionItems,
-		} );
-
-		wrapper.findComponent( ValueTypeDropDown ).vm.$emit( 'input', optionItems.Matching );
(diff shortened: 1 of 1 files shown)


tests/unit/components/ValueTypeDropDown.spec.ts
```diff
@@ -8,15 +8,10 @@ describe( 'ValueTypeDropDown.vue', () => {
 		const optionItems = PropertyValueRelation;
 		const wrapper = shallowMount( ValueTypeDropDown );
 
-		wrapper.setData( {
-			selected: optionItems.Matching,
-			optionItems: optionItems,
-		} );
-
-		wrapper.findComponent( ValueTypeDropDown ).vm.$emit( 'input', optionItems.Matching );
+		wrapper.find(`option[value="${optionItems.Regardless}"]`).setSelected();
```
guergana: It is failing on the linter. Other than that looks good. :+1:

guergana (changes requested):

tests/unit/components/ValueTypeDropDown.spec.ts
micgro42: Thanks, fixed it!

guergana (approved):
looks great! :100:

PR #78, opened 2020-11-24
Make the Query Builder configurable at build time

bereket-WMDE:
[T267671](https://phabricator.wikimedia.org/T267671)
just an initial draft commit to bounce of some ideas
main problem is, we have different configurations (tooforge and docker) and we need to find a cleaner way to set up
env variables for both situations

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,6 +1,9 @@
 # The port for the development server
 PORT=8080
 
+VUE_APP_WIKIBASE_API_URL='https://www.wikidata.org/w/api.php'
+VUE_APP_QUERY_SERVICE_EMBED_URL='https://query.wikidata.org/embed.html'
+
 #Saucelabs access data - needed if testing from local-cli
 SAUCE_USERNAME='<username>'
 SAUCE_ACCESS_KEY='<access_key>'
--- a/.env.production
+++ b/.env.production
@@ -1,0 +1,2 @@
+VUE_APP_WIKIBASE_API_URL='https://www.wikidata.org/w/api.php'
+VUE_APP_QUERY_SERVICE_EMBED_URL='https://query.wikidata.org/embed.html'
--- a/build/updater.sh
+++ b/build/updater.sh
@@ -2,6 +2,7 @@
 cd /data/project/query-builder-test/query-builder
 git reset --hard origin/master
 git pull
+export NODE_ENV=production
 npm install npm@6.14.8
 ./node_modules/.bin/npm install
 ./node_modules/.bin/npm run build
--- a/src/components/QueryResult.vue
+++ b/src/components/QueryResult.vue
@@ -9,10 +9,9 @@
 		<div v-else-if="encodedQuery.length === 0" class="querybuilder__result__description">
 			{{ $i18n('query-builder-result-placeholder')}}
 		</div>
-		<!-- TODO use global config variables instead of hardcoding the url-->
 		<iframe
 			v-else
-			:src="'https://query.wikidata.org/embed.html#' + encodedQuery"
+			:src="queryServiceEmbedUrl + '#' + encodedQuery"
 			:key="iframeRenderKey"
 			class="querybuilder__result__iframe"
 			referrerpolicy="origin"
@@ -28,6 +27,11 @@ import { mapState } from 'vuex';
 
 export default Vue.extend( {
 	name: 'QueryResult',
+	data() {
+		return {
+			queryServiceEmbedUrl: process.env.VUE_APP_QUERY_SERVICE_EMBED_URL,
+		};
+	},
 	props: {
 		encodedQuery: {
 			type: String,
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -16,3 +16,5 @@ services:
         command: 'npm run serve'
         ports:
             - '${PORT}:${PORT}'
+        environment:
+            - NODE_ENV=development
--- a/src/main.ts
+++ b/src/main.ts
@@ -9,7 +9,7 @@ Vue.config.productionTip = false;
 const services = new QueryBuilderServices();
 services.set( 'searchEntityRepository', new FetchSearchEntityRepository(
 	'en', // TODO: this should somehow depend on the browser via T263553
-	'https://www.wikidata.org/w/api.php', // TODO: this should somehow be configured at build time
+	process.env.VUE_APP_WIKIBASE_API_URL || '',
 ) );
 
 new Vue( {


build/updater.sh
```diff
@@ -2,8 +2,12 @@
 cd /data/project/query-builder-test/query-builder
 git reset --hard origin/master
 git pull
+export NODE_ENV=production
 npm install npm@6.14.8
 ./node_modules/.bin/npm install
 ./node_modules/.bin/npm run build
 rm -rf public_html/*
 cp -R dist/* ../public_html/
+cat <<EOT >> env.production.local
```
micgro42: Isn't this part needed before `./node_modules/.bin/npm run build`?

I feel like we could have one file `env.production` and that holds our production configuration. It would probably be read with `source env.production` or something similar?

Maybe at some later point we can enable a local override, when 3rd parties also want to use it themselves or we have some sensitive data here that shouldn't be in public version control.

micgro42 (commented):
I like the direction. We probably want different approaches for different variables.

`VUE_APP_WIKIDATA_API_URL`:
* I think this should be `WIKIBASE_API_URL`, we probably don't need to specify that it is a vue app and it could be used for other Wikibase instances than just Wikidata
* It is probably ok if that defaults to Wikidata, we can then override in the Browser Tests and locally, if we want

`QUERY_SERVICE_EMBED_URL="https://query.wikidata.org/embed.html"`:
* This can also default to the Wikidata Query service, as above

Sauce Labs:
* It is probably ok if that stays empty by default, it will only be set via the GitHub CI interface

Future tracking:
* When we get to it, we will probably keep it empty by default and only override that for production

bereket-WMDE:
> I like the direction. We probably want different approaches for different variables.
> 
> `VUE_APP_WIKIDATA_API_URL`:
> 
>     * I think this should be `WIKIBASE_API_URL`, we probably don't need to specify that it is a vue app and it could be used for other Wikibase instances than just Wikidata
> 
>     * It is probably ok if that defaults to Wikidata, we can then override in the Browser Tests and locally, if we want
> 
> 
> `QUERY_SERVICE_EMBED_URL="https://query.wikidata.org/embed.html"`:
> 
>     * This can also default to the Wikidata Query service, as above
> 
> 
> Sauce Labs:
> 
>     * It is probably ok if that stays empty by default, it will only be set via the GitHub CI interface
> 
> 
> Future tracking:
> 
>     * When we get to it, we will probably keep it empty by default and only override that for production


regarding 'VUE_APP_WIKIDATA_API_URL'. we need the prefix 'VUE_APP', so vuejs can pick it up as an env variable.

build/updater.sh
bereket-WMDE: @micgro42  regarding the naming : (VUE_APP_WIKIDATA_API_URL), vue doesn't pick up env files unless they're prefixed with VUE_APP, as noted [here](https://cli.vuejs.org/guide/mode-and-env.html#environment-variables)

bereket-WMDE:
@micgro42 added some new changes

build/updater.sh
micgro42: > @micgro42 regarding the naming : (VUE_APP_WIKIDATA_API_URL), vue doesn't pick up env files unless they're prefixed with VUE_APP, as noted [here](https://cli.vuejs.org/guide/mode-and-env.html#environment-variables)

Ok, that makes sense. Then yes, let's prefix the names with that.

.env.example
```diff
@@ -1,6 +1,9 @@
 # The port for the development server
 PORT=8080
 
+VUE_APP_WIKIBASE_API_URL=
```
micgro42: If it is an example, then it should have example content :)

Using the same as in .env.production is fine.

.env.production
```diff
@@ -0,0 +1,9 @@
+# The port for the development server
+PORT=8080
+
+VUE_APP_WIKIBASE_API_URL='https://www.wikidata.org/w/api.php'
+VUE_APP_QUERY_SERVICE_EMBED_URL='https://query.wikidata.org/embed.html#'
```
micgro42: This should not have the trailing `#`. That is conceptually part of the hash that we appending in code

.env.production
```diff
@@ -0,0 +1,9 @@
...
+
+#Saucelabs access data - needed if testing from local-cli
+SAUCE_USERNAME='<username>'
```
micgro42: I think we can drop these `SAUCE_` keys completely from this file. We're not gaining anything from them being here with non-data.

build/updater.sh
```diff
@@ -2,8 +2,10 @@
 cd /data/project/query-builder-test/query-builder
 git reset --hard origin/master
 git pull
+export NODE_ENV=production
 npm install npm@6.14.8
 ./node_modules/.bin/npm install
+source env.production
```
micgro42: From the page that you linked above: is it possible that env.production is picked up automatically in the build step?

Could you verify that? If so, then we don't need that line here after all.

src/components/QueryResult.vue
```diff
@@ -28,6 +28,12 @@ import { mapState } from 'vuex';
 
 export default Vue.extend( {
 	name: 'QueryResult',
+	data() {
+		return {
+			queryServiceEmbedUrl: process.env.VUE_APP_QUERY_SERVICE_EMBED_URL ||
+				'https://query.wikidata.org/embed.html#',
```
micgro42: The hash should not be part of this string here as well

src/components/QueryResult.vue
```diff
@@ -12,7 +12,7 @@
 		<!-- TODO use global config variables instead of hardcoding the url-->
 		<iframe
 			v-else
-			:src="'https://query.wikidata.org/embed.html#' + encodedQuery"
+			:src="queryServiceEmbedUrl + encodedQuery"
```
micgro42: It is here that the hash should be added

micgro42 (commented):
Some more thoughts. Not sure yet about the fallbacks directly in the code vs. vs. just using .env vs some explicit configuration loading file, but that is maybe also something we can rework later if we see the need for it.

.env.example
bereket-WMDE: @micgro42 
In my experience, the env.example files is only supposed to show what env variables there are , and not their actual value.
especially, since its going to be part of the repo (not in .gitignore). 
what if we have secrets in the env file ? we're certainly not going to add those values in the .env.example file. 
this is all just from my experience. I'm happy to hear rebuttals.

bereket-WMDE:
> Some more thoughts. Not sure yet about the fallbacks directly in the code vs. vs. just using .env vs some explicit configuration loading file, but that is maybe also something we can rework later if we see the need for it.

@micgro42 
so the current logic is, I'm using .env and .env.prodution files. according to the [doku](https://cli.vuejs.org/guide/mode-and-env.html#environment-variables), the .env file is loaded in all cases. which can be assumed as a fallback of sorts ?
and if so, maybe we don't need to add the fallback in the code itself we can just remove them and assume .env will be automatic default

build/updater.sh
bereket-WMDE: it seems to be the case.

> Env Loading Priorities

> An env file for a specific mode (e.g. .env.production) will take higher priority than a generic one (e.g. .env).

> In addition, environment variables that already exist when Vue CLI is executed have the highest priority and will not be overwritten by .env files.

>  .env files are loaded at the start of vue-cli-service. Restart the service after making changes.

.env.example
micgro42: No, there shouldn't be any secrets in .env file. The point of showing it here is to give an example for what is expected. E.g. that the api url is `https://example.org/w/api.php` and not `https://example.org/w/index.php` or something. I.e. an example file should in principle be valid. I'm not saying that it should be a copy of the production file, just pointing out that what we currently have for those two fields in the production file would be good examples.

You might be thinking of not an `.example` file, but a `.dist` file which is a template or starter that is intended to be copied to `.local`, to be manually adjusted and then read by the application?

bereket-WMDE: I see. for now, i just copied over what we have on the env file
i was used to a different workflow. but what you said makes sense, as well

.env.production
```diff
@@ -0,0 +1,7 @@
+# The port for the development server
+PORT=8080
```
micgro42: That is already in `.env` and also unused in the production build, I think?

.env.production
```diff
@@ -0,0 +1,7 @@
...
+
+VUE_APP_ENVIROMENT="production"
```
micgro42: Where is this used?

micgro42 (commented):
> > Some more thoughts. Not sure yet about the fallbacks directly in the code vs. vs. just using .env vs some explicit configuration loading file, but that is maybe also something we can rework later if we see the need for it.
> 
> @micgro42
> so the current logic is, I'm using .env and .env.prodution files. according to the [doku](https://cli.vuejs.org/guide/mode-and-env.html#environment-variables), the .env file is loaded in all cases. which can be assumed as a fallback of sorts ?
> and if so, maybe we don't need to add the fallback in the code itself we can just remove them and assume .env will be automatic default

Mh, if the .env file is always loaded, then it might make sense to move the `VUE_APP_WIKIBASE_API_URL` and `VUE_APP_QUERY_SERVICE_EMBED_URL` there as well, as we usually want to use those results during local development. Though, we should keep the `env.production` file, because we _do not_ want to use the same tracking configuration for local development and production.
What do you think?

.env.production
bereket-WMDE: no where :) good catch. will remove it

.env.production
bereket-WMDE: yap. will be removed

bereket-WMDE:
> > > Some more thoughts. Not sure yet about the fallbacks directly in the code vs. vs. just using .env vs some explicit configuration loading file, but that is maybe also something we can rework later if we see the need for it.
> > 
> > 
> > @micgro42
> > so the current logic is, I'm using .env and .env.prodution files. according to the [doku](https://cli.vuejs.org/guide/mode-and-env.html#environment-variables), the .env file is loaded in all cases. which can be assumed as a fallback of sorts ?
> > and if so, maybe we don't need to add the fallback in the code itself we can just remove them and assume .env will be automatic default
> 
> Mh, if the .env file is always loaded, then it might make sense to move the `VUE_APP_WIKIBASE_API_URL` and `VUE_APP_QUERY_SERVICE_EMBED_URL` there as well, as we usually want to use those results during local development. Though, we should keep the `env.production` file, because we _do not_ want to use the same tracking configuration for local development and production.
> What do you think?

do you mean we should remove those two variables from the .env.production file ?

src/components/QueryResult.vue
```diff
@@ -12,7 +12,7 @@
 		<!-- TODO use global config variables instead of hardcoding the url-->
```
micgro42: This comment can now be removed :)

micgro42:
> > Mh, if the .env file is always loaded, then it might make sense to move the `VUE_APP_WIKIBASE_API_URL` and `VUE_APP_QUERY_SERVICE_EMBED_URL` there as well, as we usually want to use those results during local development. Though, we should keep the `env.production` file, because we _do not_ want to use the same tracking configuration for local development and production.
> > What do you think?
> 
> do you mean we should remove those two variables from the .env.production file ?

Yes, _iff_ we want them always loaded, except in exceptions (e.g. the Browser tests), then it makes sense to treat them as the default and that would mean to have them in the `.env` file and not the `.env.production` file.

bereket-WMDE:
> > > Mh, if the .env file is always loaded, then it might make sense to move the `VUE_APP_WIKIBASE_API_URL` and `VUE_APP_QUERY_SERVICE_EMBED_URL` there as well, as we usually want to use those results during local development. Though, we should keep the `env.production` file, because we _do not_ want to use the same tracking configuration for local development and production.
> > > What do you think?
> > 
> > 
> > do you mean we should remove those two variables from the .env.production file ?
> 
> Yes, _iff_ we want them always loaded, except in exceptions (e.g. the Browser tests), then it makes sense to treat them as the default and that would mean to have them in the `.env` file and not the `.env.production` file.

right. buuuut,i think we'll need them in the env.production file as well because we are using those variables in toolorge (which is set to  NODE_ENV=production). and since the .env file is not checked into the codebase, we'll need the .env.production

build/updater.sh
```diff
@@ -2,6 +2,7 @@
 cd /data/project/query-builder-test/query-builder
 git reset --hard origin/master
 git pull
+export NODE_ENV=production
 npm install npm@6.14.8
```
Ladsgroup, line 6: I assume we need to inject values of .env.production here too. Right? or do I need more coffee?

bereket-WMDE: lol no :)
once the NODE_ENV is set to production, Vue will automatically pick up the values from .env.production file

Ladsgroup (approved):

PR #79, opened 2020-11-25
Start automatically disabling value for partially supported properties

Ladsgroup:
Bug: T268447

### DIFF
--- a/src/components/QueryBuilder.vue
+++ b/src/components/QueryBuilder.vue
@@ -17,6 +17,7 @@
 				/>
 				<ValueTypeDropDown
 					v-model="selectedPropertyValueRelation"
+					:disabled="limitedSupport"
 				/>
 				<TextInput
 					class="querybuilder__rule__value"
@@ -69,6 +70,7 @@ export default Vue.extend( {
 				value: null as null | Error,
 			},
 			selectedOption: '',
+			limitedSupport: false,
 			propertyValueRelation: PropertyValueRelation,
 		};
 	},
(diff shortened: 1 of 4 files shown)


src/components/QueryBuilder.vue
```diff
@@ -85,6 +87,20 @@ export default Vue.extend( {
 			this.$store.dispatch( 'setErrors', validationResult.formErrors );
 			this.fieldErrors = validationResult.fieldErrors;
 		},
+		validateForLimitedSupport( selectedProperty: SearchResult ): void {
+			this.limitedSupport = false;
+			this.fieldErrors.property = null;
+			this.selectedPropertyValueRelation = PropertyValueRelation.Matching;
+			const allowedDatatypes = [ 'external-id', 'string' ];
+			if ( selectedProperty && allowedDatatypes.indexOf( selectedProperty.datatype ) === -1 ) {
```
micgro42: This can use [Array.prototype.includes()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes)

src/data-access/FetchSearchEntityRepository.ts
```diff
@@ -47,6 +47,48 @@ export default class FetchSearchEntityRepository implements SearchEntityReposito
 
 		const data = await response.json();
 
+		// Inject data types
```
micgro42: I'm sorry I have to say this after all this work is already done, but isn't the datatype included in the `wbsearchentities` response?

See for example https://www.wikidata.org/w/api.php?action=wbsearchentities&search=commons&language=en&type=property

micgro42 (changes requested):

src/data-access/FetchSearchEntityRepository.ts
Ladsgroup: How on earth I missed it? *Headdesks*

src/components/QueryBuilder.vue
Ladsgroup: Done.

micgro42:
Looks 90% great, but the warning gets removed when I click "Run Query".

Ladsgroup:
> Looks 90% great, but the warning gets removed when I click "Run Query".

Done now.

micgro42 (approved):

PR #80, opened 2020-11-25
Fix rootState error for propertyData being undefined

Ladsgroup:
Currently, if you try looking up a property, in the on-air moment, it
starts to emit lots of error since the property is not yet selected and
is undefined. This would fix the error.

### DIFF
--- a/src/store/getters.ts
+++ b/src/store/getters.ts
@@ -15,7 +15,7 @@ export default {
 		};
 	},
 	property( rootState: RootState ): Property | null {
-		if ( !rootState.conditionRow.propertyData.id ) {
+		if ( !rootState.conditionRow.propertyData?.id ) {
 			return null;
 		}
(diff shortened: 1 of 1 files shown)


src/store/getters.ts
```diff
@@ -15,7 +15,7 @@ export default {
 		};
 	},
 	property( rootState: RootState ): Property | null {
-		if ( !rootState.conditionRow.propertyData.id ) {
+		if ( !rootState.conditionRow.propertyData || !rootState.conditionRow.propertyData.id ) {
```
micgro42: I think we can write this nicer with optional chaining: `?.`

See https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#optional-chaining

Ladsgroup: Ooooh thanks. Changed.

micgro42 (approved):

PR #81, opened 2020-11-26
Start metrics collector

Ladsgroup:
Bug: T268635

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,7 +10,7 @@
     "e2e": "vue-cli-service test:e2e",
     "e2e:saucelabs": "env DATE=\"$(date)\" nightwatch --config nightwatch.conf.js --env sauceChrome,sauceFirefox,sauceEdge,sauceSafari",
     "serve": "vue-cli-service serve",
-    "build": "vue-cli-service build",
+    "build": "vue-cli-service build --mode production",
     "test:unit": "vue-cli-service test:unit",
     "test:lint": "vue-cli-service lint --no-fix --max-warning 0",
     "test": "npm-run-all test:* && vue-cli-service build",
--- a/src/ServicesFactory.ts
+++ b/src/ServicesFactory.ts
@@ -1,0 +1,16 @@
+import QueryBuilderServices from '@/QueryBuilderServices';
+import FetchSearchEntityRepository from '@/data-access/FetchSearchEntityRepository';
+import StatsvMetricsCollector from '@/data-access/StatsvMetricsCollector';
+
+const services = new QueryBuilderServices();
+services.set( 'searchEntityRepository', new FetchSearchEntityRepository(
+	'en', // TODO: this should somehow depend on the browser via T263553
+	process.env.VUE_APP_WIKIBASE_API_URL || '',
+) );
+
+services.set( 'metricsCollector', new StatsvMetricsCollector(
+	'Wikidata.query-builder',
+	process.env.VUE_APP_STATSV_ENDPOINT || null,
+) );
+
+export default services;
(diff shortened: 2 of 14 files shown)


micgro42 (changes requested):
Love it, but there are a couple of things left to do:

- base it on top of #78 and use the `env.production` file for configuration
- don't create that collector three times, but create it only once in the service container and use it from there (see `main.ts` and `QueryBuilderServices.ts`)
- tests are still missing :)

Ladsgroup:
> Love it, but there are a couple of things left to do:
> 
>     * base it on top of #78 and use the `env.production` file for configuration
Done
> 
>     * don't create that collector three times, but create it only once in the service container and use it from there (see `main.ts` and `QueryBuilderServices.ts`)

Done, I had to make some adjustments in general to make it work but I hope that looks okay.
> 
>     * tests are still missing :)
Added. Thanks.

package.json
```diff
@@ -10,7 +10,7 @@
     "e2e": "vue-cli-service test:e2e",
     "e2e:saucelabs": "env DATE=\"$(date)\" nightwatch --config nightwatch.conf.js --env sauceChrome,sauceFirefox,sauceEdge,sauceSafari",
     "serve": "vue-cli-service serve",
-    "build": "vue-cli-service build",
+    "build": "vue-cli-service build --mode production",
```
micgro42, line 13: Is that not the default for build?

src/store/actions.ts
```diff
@@ -1,17 +1,17 @@
...
 import SearchResult from '@/data-access/SearchResult';
 import Error from '@/data-model/Error';
 import Property from '@/data-model/Property';
 import PropertyValueRelation from '@/data-model/PropertyValueRelation';
+import QueryBuilderServices from '@/QueryBuilderServices';
 
 // eslint-disable-next-line max-len
 // eslint-disable-next-line @typescript-eslint/explicit-function-return-type, @typescript-eslint/explicit-module-boundary-types
-export default ( searchEntityRepository: SearchEntityRepository ) => ( {
+export default ( services: QueryBuilderServices ) => ( {
```
micgro42: No, this replaces the dependency injection with the service locator anti-pattern. Rather, this should inject both services individually here.

Currently, this may feel a bit pointless, but it will make more sense as soon as we use more services in other places as well.

(However, I think it is fine as it is in App.vue)

src/ServicesFactory.ts
```diff
@@ -0,0 +1,16 @@
+import QueryBuilderServices from '@/QueryBuilderServices';
+import FetchSearchEntityRepository from '@/data-access/FetchSearchEntityRepository';
+import StatsvMetricsCollector from '@/data-access/StatsvMetricsCollector';
+
+const services = new QueryBuilderServices();
```
micgro42, line 5: I like this ServiceFactory, thanks!

micgro42 (commented):
Almost there, but I'm not sure if the change to how we inject the services into the actions is ideal

package.json
Ladsgroup: I like it to be explicit

src/store/actions.ts
Ladsgroup: hmm, makes sense. Fixed now.

src/ServicesFactory.ts
Ladsgroup: I just copied the logic from mediawiki mwhahahaha

micgro42 (approved):
Nice, thank you !

Ladsgroup:
> Nice, thank you !

Thanks!

Ball

PR #82, opened 2020-11-26
Created list of allowed datatypes

guergana:
Not sure about the location. Put it in the root folder for now. Open to suggestions.

### DIFF
--- a/src/allowedDataTypes.ts
+++ b/src/allowedDataTypes.ts
@@ -1,0 +1,24 @@
+const allowedDataTypes = [
+	/* string typed datatypes */
+	'commonsMedia',
+	'geo-shape',
+	'string',
+	'tabular-data',
+	'url',
+	'external-id',
+	'musical-notation',
+	'math',
+	/* other (non-string) typed datatypes - TODO: add support
+    'time',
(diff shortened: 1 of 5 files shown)


src/allowedDataTypes.ts
```diff
@@ -0,0 +1,21 @@
+const allowedDataTypes = [
```
Ladsgroup, line 1: I'm pretty sure we don't yet support around of these data types like time or wikibase-item, they all need to be removed and only things that have "value type" of string should stay. I will find the list for you now.

Ladsgroup (changes requested):

src/allowedDataTypes.ts
Ladsgroup: The list: https://github.com/wikimedia/Wikibase/blob/master/lib/WikibaseLib.datatypes.php#L18

guergana: oh, ok. That was not made clear in the ticket, I can leave them commented out to be added later easily.

Ladsgroup (approved):
Thanks.

PR #83, opened 2020-11-26
Fix banana tests

Ladsgroup:
Translatewiki bot adds "@metadata" key in qqq.json that doesn't exist in
en.json and the test mistakenly takes it as an orphan description and
error out. Ignoring that particular key.

### DIFF
--- a/tests/unit/i18n/banana.spec.ts
+++ b/tests/unit/i18n/banana.spec.ts
@@ -27,6 +27,9 @@ describe( 'banana tests', () => {
 		}
 
 		for ( messageKey in messages.qqq ) {
+			if ( messageKey === '@metadata' ) {
+				continue;
+			}
 			expect( messages.en[ messageKey ] ).toBeTruthy();
 		}
 	} );


bereket-WMDE:
> Translatewiki bot adds "@metadata" key in qqq.json that doesn't exist in
> en.json and the test mistakenly takes it as an orphan description and
> error out. Ignoring that particular key.

@Ladsgroup is it only in the en.json file ?
if so, what is the reason ?

Ladsgroup:
> > Translatewiki bot adds "@metadata" key in qqq.json that doesn't exist in
> > en.json and the test mistakenly takes it as an orphan description and
> > error out. Ignoring that particular key.
> 
> @Ladsgroup is it only in the en.json file ?
> if so, what is the reason ?

the metadata gets added to anything but en.json because they get edited in translatewiki (including qqq) and they need to credit the authors. Here's an example: https://github.com/wmde/query-builder/blob/master/public/i18n/fa.json

micgro42 (approved):

PR #84, opened 2020-11-27
Add execute right to updater.sh

Ladsgroup:
Without it, the pods fail with this error:
tools.query-builder-test@tools-sgebastion-07:~/query-builder$ kubectl logs test-6l8kv
/bin/sh: 1: /data/project/query-builder-test/query-builder/build/updater.sh: Permission denied


bereket-WMDE:
the failing checks should be fixed in [85](https://github.com/wmde/query-builder/pull/85)

bereket-WMDE (approved):

PR #87, opened 2020-11-30
Fix for dropdown always changing to matching on validation

guergana:


Ladsgroup (changes requested):
I added that line for another edge case, it's when you use a property with limited support but then change it to another property with full support, without this line, the fields stay disabled. This should fix it in another way.

guergana:
> I added that line for another edge case, it's when you use a property with limited support but then change it to another property with full support, without this line, the fields stay disabled. This should fix it in another way.

I think the logic of this function needs to be re-thought because right now it's asking to change to `matching` whenever the property value has a value, which is not correct. This is not taking into account that there is a `without` case.

guergana (commented):
Looks good.

Ladsgroup (approved):

PR #88, opened 2020-11-30
Add CI-Job to track pull request code coverage

micgro42:
This (https://github.com/mattallty/jest-github-action) seems to be the most-used action that does not depend on an external service.

This has only a "run on changed code only" option, not sure if would benefit from that.

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -1,0 +1,24 @@
+name: Jest Code Coverage
+
+on: pull_request
+
+jobs:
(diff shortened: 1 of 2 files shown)


micgro42:
Meh. Maybe, I'll go with sonar after all.

Action: https://github.com/marketplace/actions/sonarcloud-scan
related thread: https://community.sonarsource.com/t/code-coverage-doesnt-work-with-github-action/16747

micgro42:
superseded by #176

PR #89, opened 2020-11-30
[PoC] Run browser tests against fake API

micgro42:
These browser tests are now running in CI! It will take some work to make them runnable locally as well.

Besides some cleanup, the next step would be:

- [ ] figure out for which scenarios to run the tests - we don't want too many, because flakyness
- [ ] remove linting and maybe unittests in the browser test job -- we have a dedicated job for those
- [ ] make it possible to run the browser tests locally


Ladsgroup:
Since we don't care about IE11 anymore, can we use something else instead of sauchlabs? :pleading_face:

micgro42:
> Since we don't care about IE11 anymore, can we use something else instead of sauchlabs? pleading_face

Yeah, we should talk about that. Though I do not know the alternatives well enough to make an educated recommendation.

micgro42:
Reading a bit more about what kind of backend netlify offers, this should be relatively straight-forward to implement with [Netlify functions](https://kentcdodds.com/blog/super-simple-start-to-netlify-functions) (i.e. serverless + AWS under the hood). However, that would mean that we loose Netlify as a pre-merge testing-ground.

Not sure what is more useful to have

jakobw (changes requested):
The main problem I'm seeing addressed here are the faked network interactions with the API and the query results iframe. Netlify serverless functions are a pretty cool solution, but it makes me a little uncomfortable to lock us into this service here. I think there are some decent alternatives out there which would achieve the same thing without this side effect.

The [decision for Wikit to go with Nightwatch and saucelabs](https://wmde.github.io/wikit/?path=/story/documentation-decisions-and-adrs-adrs-10-browser-testing--page) was mainly driven by the requirement to cover IE11. Without this requirement, Cypress would likely have become the tool of choice, which brings the functionality for [stubbing network requests](https://docs.cypress.io/guides/guides/network-requests.html#Testing-Strategies) out of the box.

We could also keep Nightwatch, and stub the requests with another library. I've recently used [Mirage JS](https://miragejs.com/) to intercept http requests on the client side, which worked really well. [Stackoverflow claims that nock could work as well](https://stackoverflow.com/questions/38404013/nightwatch-mock-http-requests).

Both of these alternatives would also enable you to easily run the browser tests locally, too.

micgro42:
> The main problem I'm seeing addressed here are the faked network interactions with the API and the query results iframe. Netlify serverless functions are a pretty cool solution, but it makes me a little uncomfortable to lock us into this service here. I think there are some decent alternatives out there which would achieve the same thing without this side effect.
> 
> The [decision for Wikit to go with Nightwatch and saucelabs](https://wmde.github.io/wikit/?path=/story/documentation-decisions-and-adrs-adrs-10-browser-testing--page) was mainly driven by the requirement to cover IE11. Without this requirement, Cypress would likely have become the tool of choice, which brings the functionality for [stubbing network requests](https://docs.cypress.io/guides/guides/network-requests.html#Testing-Strategies) out of the box.

Yes, after we dropped the IE requirement, Cypress was the favorite of the engineers, but engineering management decided for us to use SauceLabs unless we have a reason not to.
 
> We could also keep Nightwatch, and stub the requests with another library. I've recently used [Mirage JS](https://miragejs.com/) to intercept http requests on the client side, which worked really well. [Stackoverflow claims that nock could work as well](https://stackoverflow.com/questions/38404013/nightwatch-mock-http-requests).
> 
> Both of these alternatives would also enable you to easily run the browser tests locally, too.

I will look into those, thanks!


As to my original question: An thoughts on what should be the guiding principle for what to cover with browser tests in light of their significant trade-offs?
My current line of thinking was to have a very small number of tests only. Maybe one test that adds conditions of every input type and also rotates through the value relation options, negation and condition relation options. Maybe another to toggle the two global options?

micgro42:
superseded by #177

jakobw:
> Yes, after we dropped the IE requirement, Cypress was the favorite of the engineers, but engineering management decided for us to use SauceLabs unless we have a reason not to.

Oh, I didn't know that. @itamargiv and I worked on #177 as a fun learning GB&C project and didn't necessarily intend for it to become anything official. We didn't ask EMs for permission either. The request stubbing worked beautifully with Cypress, but I don't know if this is reason enough to drop nightwatch and saucelabs. I'd think it's worth a look, though.

> As to my original question: An thoughts on what should be the guiding principle for what to cover with browser tests in light of their significant trade-offs?
My current line of thinking was to have a very small number of tests only. Maybe one test that adds conditions of every input type and also rotates through the value relation options, negation and condition relation options. Maybe another to toggle the two global options?

Good question, and we also pondered that for a few minutes yesterday. A small number of tests, and at least one for each supported data type sounds reasonable. I think it would be good to go through the whole query building process for each of these, and then inspect the resulting query as we did here 2cfc02268f30009e47815df5660e448301b9427a. That covers both the interactive elements and the actual query building logic from end to end, but without relying on the API or query results.

PR #90, opened 2020-12-01
Modified state to accept multiple conditions

guergana:
+ multiple values (first commit)
+ multiple properties (second commit)
+ multiple PropertyValueRelations (third commit)

### DIFF
--- a/src/store/getters.ts
+++ b/src/store/getters.ts
@@ -27,8 +27,10 @@ export default {
 			return rootState.conditionRows[ conditionIndex ].valueData.value;
 		};
 	},
-	propertyValueRelation( rootState: RootState ): PropertyValueRelation {
-		return rootState.conditionRows[ 0 ].propertyValueRelationData.value;
+	propertyValueRelation( rootState: RootState ) {
+		return ( conditionIndex: number ): PropertyValueRelation => {
+			return rootState.conditionRows[ conditionIndex ].propertyValueRelationData.value;
+		};
 	},
 	getErrors( rootState: RootState ): Error[] {
 		return rootState.errors;
--- a/src/store/getters.ts
+++ b/src/store/getters.ts
@@ -8,23 +8,26 @@ export default {
 	query( rootState: RootState ): QueryRepresentation {
 		return {
 			condition: {
-				propertyId: rootState.conditionRow.propertyData.id,
-				value: rootState.conditionRow.valueData.value,
-				propertyValueRelation: rootState.conditionRow.propertyValueRelationData.value,
+				propertyId: rootState.conditionRows[ 0 ].propertyData.id,
+				value: rootState.conditionRows[ 0 ].valueData.value,
+				propertyValueRelation: rootState.conditionRows[ 0 ].propertyValueRelationData.value,
 			},
 		};
 	},
 	property( rootState: RootState ): Property | null {
-		if ( !rootState.conditionRow.propertyData?.id ) {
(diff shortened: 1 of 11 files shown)


src/store/getters.ts
```diff
@@ -7,24 +7,28 @@ import PropertyValueRelation from '@/data-model/PropertyValueRelation';
 export default {
 	query( rootState: RootState ): QueryRepresentation {
 		return {
-			condition: {
-				propertyId: rootState.conditionRow.propertyData.id,
-				value: rootState.conditionRow.valueData.value,
-				propertyValueRelation: rootState.conditionRow.propertyValueRelationData.value,
-			},
+			conditions: [ {
```
micgro42: I think for this PR it is enough to still just return a single condition here. While we have to extend QueryRepresentation to take multiple conditions at some point, doing so here makes this change to big.

src/sparql/QueryRepresentation.ts
```diff
@@ -7,5 +7,5 @@ export type Condition = {
 };
 
 export default interface QueryRepresentation {
-	condition: Condition;
+	conditions: Condition[];
```
micgro42: I think for the purpose of this PR it is ok if this stays a single condition for now. We'll extend that in a follow-up PR

micgro42 (commented):
This is pretty solid work so far! It still needs the respective adjustment to be applied to the propertyValueRelation and removing the changes to the QueryRepresentation etc., but than this is should be almost good to go :+1:

tests/unit/components/QueryBuilder.spec.ts
```diff
@@ -40,43 +47,45 @@ describe( 'QueryBuilder.vue', () => {
...
 	it( 'passes the selected property back to the PropertyLookup', async () => {
 		const property = { label: 'postal code', id: 'P123' };
-		const wrapper = shallowMount( QueryBuilder, {
-			store: newStore( {
-				property: jest.fn().mockReturnValue( property ),
-			} ),
-			localVue,
-		} );
+		const conditionIndex = 0;
+		const wrapper = shallowMount( QueryBuilder, { store: newStore(), localVue } );
```
micgro42: We probably want to explicitly inject the property getter here (can be very simple like `() => () => ({ label: 'postal code', id: 'P123' })`), so that if the dummy code in the default store is changed, this test won't suddenly start to fail. As it is currently, this test has a hidden dependency on the mock data in the `newStore()` method.

tests/unit/components/QueryBuilder.spec.ts
```diff
@@ -40,43 +47,45 @@ describe( 'QueryBuilder.vue', () => {
...
 		const store = newStore();
+		const conditionIndex = 0;
 		store.dispatch = jest.fn();
 		const wrapper = shallowMount( QueryBuilder, {
 			store,
 			localVue,
 		} );
 
-		wrapper.findComponent( PropertyLookup ).vm.$emit( 'input', property );
+		wrapper.findAllComponents( PropertyLookup ).at( conditionIndex ).vm.$emit( 'input', property );
```
micgro42, line 66: Nice!

micgro42 (changes requested):
I have one tiny detail, then it is ready to be merged. All the other code is fine! :D

tests/unit/components/QueryBuilder.spec.ts
micgro42: Done

micgro42 (approved):

PR #91, opened 2020-12-02
Split QueryCondition component out of QueryBuilder

Ladsgroup:
Bug: T269133

### DIFF
--- a/src/components/QueryBuilder.vue
+++ b/src/components/QueryBuilder.vue
@@ -10,7 +10,8 @@
 		<div role="form">
 			<h2 class="querybuilder__find-title"
 				v-i18n="{msg: 'query-builder-find-all-items'}" />
-			<QueryCondition/>
+			<QueryCondition
+				ref="condition"/>
 			<div class="querybuilder__run">
 				<Button
 					@click.native="runQuery"
@@ -26,7 +27,7 @@
 </template>
 
 <script lang="ts">
-import Vue from 'vue';
+import Vue, { VueConstructor } from 'vue';
 import { mapState } from 'vuex';
 import { Button } from '@wmde/wikit-vue-components';
 
@@ -35,7 +36,7 @@ import QueryResult from '@/components/QueryResult.vue';
 import buildQuery from '@/sparql/buildQuery';
 import Validator from '@/form/Validator';
 
-export default Vue.extend( {
+export default ( Vue as VueConstructor<Vue & { $refs: { condition: InstanceType<typeof QueryCondition> } }> ).extend( {
 	name: 'QueryBuilder',
 	data() {
 		return {
@@ -61,6 +62,8 @@ export default Vue.extend( {
 			const validationResult = validator.validate();
 			this.errors = validationResult.formErrors;
 			this.$store.dispatch( 'setErrors', validationResult.formErrors );
+
+			this.$refs.condition.validate();
 
 		},
 		runQuery(): void {
--- a/src/components/QueryBuilder.vue
+++ b/src/components/QueryBuilder.vue
@@ -10,26 +10,7 @@
 		<div role="form">
 			<h2 class="querybuilder__find-title"
 				v-i18n="{msg: 'query-builder-find-all-items'}" />
-			<div class="querybuilder__rule">
-				<PropertyLookup
-					v-model="selectedProperty"
-					:error="fieldErrors.property"
-				/>
-				<ValueTypeDropDown
-					v-model="selectedPropertyValueRelation"
-					:disabled="limitedSupport"
-				/>
-				<TextInput
-					class="querybuilder__rule__value"
-					:label="$i18n('query-builder-input-value-label')"
-					ref="value"
-					v-model="textInputValue"
-					:error="fieldErrors.value ?
-						{message: $i18n(fieldErrors.value.message), type: fieldErrors.value.type}: null"
-					:placeholder="$i18n('query-builder-input-value-placeholder')"
-					:disabled="selectedPropertyValueRelation === propertyValueRelation.Regardless"
-				/>
-			</div>
+			<QueryCondition/>
 			<div class="querybuilder__run">
 				<Button
 					@click.native="runQuery"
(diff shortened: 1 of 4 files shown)


src/components/QueryCondition.vue
```diff
@@ -0,0 +1,132 @@
...
+export default Vue.extend( {
+	name: 'QueryCondition',
+	data() {
+		return {
+			fieldErrors: {
+				property: null as null | Error,
+				value: null as null | Error,
+			},
+			limitedSupport: false,
+			propertyValueRelation: PropertyValueRelation,
```
micgro42: As far as I can tell, this is unused can just be dropped.

src/components/QueryCondition.vue
```diff
@@ -0,0 +1,132 @@
...
+					this.textInputValue = '';
+				}
+				this.$store.dispatch( 'updatePropertyValueRelation', selectedPropertyValueRelation );
+			},
+		},
+		textInputValue: {
+			get(): string { return this.$store.getters.value; },
+			set( value: string ): void { this.$store.dispatch( 'updateValue', value ); },
+		},
+		...mapState( {
```
micgro42: We're not using this at all in this component

src/components/QueryBuilder.vue
```diff
@@ -10,26 +10,9 @@
...
-					ref="value"
-					v-model="textInputValue"
-					:error="fieldErrors.value ?
-						{message: $i18n(fieldErrors.value.message), type: fieldErrors.value.type}: null"
-					:placeholder="$i18n('query-builder-input-value-placeholder')"
-					:disabled="selectedPropertyValueRelation === propertyValueRelation.Regardless"
-				/>
-			</div>
+			<QueryCondition
+				ref="firstCondition"
```
micgro42: Are we using this somewhere?

src/components/QueryCondition.vue
```diff
@@ -0,0 +1,132 @@
+<template>
+	<div class="querybuilder__rule" :fieldErrorsrrr="fieldErrorsrrr">
```
micgro42: `fieldErrorsrrr` seems unused?

src/components/QueryCondition.vue
```diff
@@ -0,0 +1,132 @@
...
+			get(): string { return this.$store.getters.value; },
+			set( value: string ): void { this.$store.dispatch( 'updateValue', value ); },
+		},
+		...mapState( {
+			errors: 'errors',
+		} ),
+	},
+	watch: {
+		'$store.state.errors': function (): void {
+			this.validate();
```
micgro42: Do we actually need this? What does it do?

micgro42 (changes requested):
Looking good, but I think there are still some things left to finish

src/components/QueryCondition.vue
Ladsgroup: Done

src/components/QueryCondition.vue
Ladsgroup: Done

src/components/QueryBuilder.vue
Ladsgroup: No, I put it for later but we can add it later too

src/components/QueryCondition.vue
Ladsgroup: Oh it was leftover from debug.

src/components/QueryCondition.vue
Ladsgroup: Yeah we need this. It basically adds the per-field validation messages when the query is incomplete. We should rework it once the state is in a better shape.

src/components/QueryCondition.vue
Ladsgroup: It is used in the template: propertyValueRelation.Regardless, when I removed it, lots of tests failed with:
 ● QueryCondition.vue › updates the store value when the user fills in the value textfield

    TypeError: Cannot read property 'Regardless' of undefined

micgro42: That should be `PropertyValueRelation.Regardless` (capital `P`, the thing that is imported)

Ladsgroup: That doesn't work, I think inside template, you can't do it. I can extract it to a method and use that instead.

src/components/QueryCondition.vue
micgro42: I see. That works, but is quite magical. Going with the ref after all is maybe the better idea. I.e. having a ref `condition` on the `QueryCondition` in the `QueryBuilder` component and then calling `condition.validate()` in the `QueryBuilder`'s `validate` method.

You may need the following code for giving the correct type to the ref: https://github.com/wmde/wikit/commit/bad24a9eba4a2be4fec9539d973c6636381bd212#diff-c000023c1d09de1efda5d372b6a61658f3158308cbb6d585822d328ac512389eR55

src/components/QueryBuilder.vue
```diff
@@ -82,31 +51,17 @@ export default Vue.extend( {
 			this.$store.dispatch( 'incrementMetric', metric );
 		},
 		validate(): void {
+			// TODO: Clean up FormValue <-> Query relation
```
micgro42, line 55: Maybe we can transform this into a getter `isQueryBuilderValid`? Then we can look at its result in this method and set the state accordingly. But that can be done in a follow-up

src/components/QueryCondition.vue
micgro42: > I can extract it to a method and use that instead.

Yes, a dedicated Method determining whether the Dropdown is disabled seems less confusing than a state that is actually just used as an enum.

src/components/QueryBuilder.vue
Ladsgroup: Yes, This PR is already getting pretty big.

src/components/QueryCondition.vue
Ladsgroup: Done

src/components/QueryCondition.vue
Ladsgroup: Done

micgro42: Yay :)

src/components/QueryCondition.vue
micgro42: :+1:

micgro42 (approved):

PR #92, opened 2020-12-03
create AddCondition button

guergana:
first commit: 

- added text for wikitranslate
- added english translation

second commit:

- created AddCondition button
- added test to verify proper implementation of emitter

third commit:

- Use AddCondition in QueryBuilder component

### DIFF
--- a/src/components/QueryBuilder.vue
+++ b/src/components/QueryBuilder.vue
@@ -30,6 +30,7 @@
 					:disabled="selectedPropertyValueRelation === propertyValueRelation.Regardless"
 				/>
 			</div>
+			<AddCondition />
 			<div class="querybuilder__run">
 				<Button
 					@click.native="runQuery"
@@ -58,6 +59,7 @@ import Error from '@/data-model/Error';
 import buildQuery from '@/sparql/buildQuery';
 import Validator from '@/form/Validator';
 import allowedDatatypes from '@/allowedDataTypes';
+import AddCondition from '@/components/AddCondition.vue';
 
 export default Vue.extend( {
 	name: 'QueryBuilder',
@@ -159,6 +161,7 @@ export default Vue.extend( {
 		QueryResult,
 		PropertyLookup,
 		ValueTypeDropDown,
+		AddCondition,
 	},
 } );
 </script>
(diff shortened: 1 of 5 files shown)


Ladsgroup (dismissed):

src/components/QueryBuilder.vue
```diff
@@ -222,4 +227,9 @@ a {
 .querybuilder__run {
 	margin-block-start: $dimension-layout-medium;
 }
+
+.querybuilder__add-condition {
+	float: inline-end;
```
guergana: do we want to use `flex` here instead? in that case we need a flex wrapper only for this element

src/components/QueryBuilder.vue
```diff
@@ -30,6 +30,9 @@
 					:disabled="selectedPropertyValueRelation === propertyValueRelation.Regardless"
 				/>
 			</div>
+			<div class="querybuilder__add-condition">
```
Ladsgroup: I think this can just go inside the component. I think it makes more sense there.

guergana: Done

src/components/QueryBuilder.vue
bereket-WMDE: seems like it has terrible browser support.
https://caniuse.com/?search=inline-end

bereket-WMDE (commented):
it looks good to me

src/components/QueryBuilder.vue
micgro42: I think our postcss is supposed to take care of that, but somehow this is not working. I.e. it is broken in chrome.

micgro42: On inspecting it further, that part of postcss seems to correctly replace it with `float: right` and `float: left`, but we are not picking it up because there is still some confusion about the `dir` attribute. I'll look.

micgro42 (approved):
This works as intended. Postcss produces the following styles:

```css
[dir=ltr] .querybuilder__add-condition[data-v-4ec7d228]{float:right}
[dir=rtl] .querybuilder__add-condition[data-v-4ec7d228]{float:left}
.querybuilder__add-condition[data-v-4ec7d228]{float:inline-end}
```

That is what we want. The problem is that we are currently always using `[dir=auto]` and so those styles don't match. However, that problem is independent of this PR and so I'm approving this.

PR #93, opened 2020-12-03
Load more results for PropertyLookup on scroll

bereket-WMDE:
Bug: T268508

### DIFF
--- a/src/components/PropertyLookup.vue
+++ b/src/components/PropertyLookup.vue
@@ -7,6 +7,7 @@
 		:search-input.sync="search"
 		:placeholder="$i18n('query-builder-property-lookup-placeholder')"
 		:label="$i18n('query-builder-property-lookup-label')"
+		v-on:scroll="handleScroll"
 	>
 		<template
 			v-slot:no-results
@@ -21,6 +22,7 @@ import Vue, { PropType } from 'vue';
 
 import { Lookup } from '@wmde/wikit-vue-components';
 import SearchResult from '@/data-access/SearchResult';
+import SearchOptions from '@/data-access/SearchOptions';
 
 export default Vue.extend( {
 	name: 'PropertyLookup',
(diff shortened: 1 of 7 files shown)


src/components/PropertyLookup.vue
```diff
@@ -7,6 +7,7 @@
 		:search-input.sync="search"
 		:placeholder="$i18n('query-builder-property-lookup-placeholder')"
 		:label="$i18n('query-builder-property-lookup-label')"
+		v-on:scroll="handelScroll"
```
Ladsgroup: typo: handle :D

tests/unit/data-access/FetchSearchEntityRepository.spec.ts
```diff
@@ -66,6 +66,7 @@ describe( 'FetchSearchEntityRepository', () => {
 			origin: '*',
 			limit,
 			offset,
```
Ladsgroup: The offset here is not needed at all: https://www.wikidata.org/w/api.php?action=help&modules=wbsearchentities

src/data-access/FetchSearchEntityRepository.ts
```diff
@@ -28,6 +28,7 @@ export default class FetchSearchEntityRepository implements SearchEntityReposito
 		}
 		if ( offset ) {
 			params.offset = `${offset}`;
```
Ladsgroup: this should be dropped

Ladsgroup (changes requested):
Thanks for doing it!

src/components/PropertyLookup.vue
bereket-WMDE: oops

tests/unit/data-access/FetchSearchEntityRepository.spec.ts
bereket-WMDE: good catch

src/data-access/FetchSearchEntityRepository.ts
bereket-WMDE: removed

Ladsgroup (approved):
LGTM now, thanks!

PR #94, opened 2020-12-04
Refactor QueryRepresentation to accept multiple conditions

Ladsgroup:
Bug: T269134

### DIFF
--- a/src/store/getters.ts
+++ b/src/store/getters.ts
@@ -7,13 +7,13 @@ import PropertyValueRelation from '@/data-model/PropertyValueRelation';
 export default {
 	query( rootState: RootState ): QueryRepresentation {
 		return {
-			conditions: [
-				{
-					propertyId: rootState.conditionRows[ 0 ].propertyData.id,
-					value: rootState.conditionRows[ 0 ].valueData.value,
-					propertyValueRelation: rootState.conditionRows[ 0 ].propertyValueRelationData.value,
-				},
-			],
+			conditions: rootState.conditionRows.map( ( condition ) => {
+				return {
+					propertyId: condition.propertyData.id,
+					value: condition.valueData.value,
+					propertyValueRelation: condition.propertyValueRelationData.value,
+				};
+			} ),
 		};
 	},
 	property( rootState: RootState ) {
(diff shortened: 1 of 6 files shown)


tests/unit/sparql/buildQuery.spec.ts
```diff
@@ -38,11 +42,34 @@ describe( 'buildQuery', () => {
...
+		expect( buildQuery( { conditions: [
+			{
+				propertyId,
+				value: '',
+				propertyValueRelation,
+			},
+		] } ) ).toEqual( `SELECT ?item WHERE { ?item (p:${propertyId}/ps:${propertyId}) _:anyValue. }` );
 	} );
 
+	it( 'builds a query from multiple conditions', () => {
```
micgro42: Yay! Can we have two more test-cases? I'm especially thinking about "matching" + "not matching" and "not matching" + "matching". As those "not matching" queries consist of two lines and I want to make sure that we get the syntax and meaning right.

src/store/getters.ts
```diff
@@ -7,11 +7,13 @@ import PropertyValueRelation from '@/data-model/PropertyValueRelation';
 export default {
 	query( rootState: RootState ): QueryRepresentation {
 		return {
-			condition: {
-				propertyId: rootState.conditionRows[ 0 ].propertyData.id,
-				value: rootState.conditionRows[ 0 ].valueData.value,
-				propertyValueRelation: rootState.conditionRows[ 0 ].propertyValueRelationData.value,
-			},
+			conditions: [
```
micgro42: I think it would be ok to now implement a `.map()` here instead of just picking the first one

tests/unit/sparql/buildQuery.spec.ts
Ladsgroup: Done

src/store/getters.ts
Ladsgroup: Done

micgro42: Thanks!

tests/unit/sparql/buildQuery.spec.ts
micgro42: Mh, I think that those are added correctly, given the other code that we have. It is just that they are broken and not actually doing what they are supposed to do.

I'll create a new bug ticket for that. That is not blocking this task

micgro42 (approved):

PR #95, opened 2020-12-04
Update to use wikit: 2.0.0-alpha.1 

Ladsgroup:
- Takes advantage of Dropdown menu
 - Set tag for property lookup search results

### DIFF
--- a/src/components/PropertyLookup.vue
+++ b/src/components/PropertyLookup.vue
@@ -17,13 +17,13 @@
 
 <script lang="ts">
 import { MenuItem } from '@wmde/wikit-vue-components/dist/components/MenuItem';
-import Vue, { PropType, VueConstructor } from 'vue';
+import Vue, { PropType } from 'vue';
 
 import { Lookup } from '@wmde/wikit-vue-components';
 import allowedDatatypes from '@/allowedDataTypes';
 import SearchResult from '@/data-access/SearchResult';
 
-export default ( Vue as VueConstructor<Vue & { $i18n: ( i: string ) => string }> ).extend( {
+export default Vue.extend( {
 	name: 'PropertyLookup',
 	components: {
 		Lookup,
--- a/src/components/PropertyLookup.vue
+++ b/src/components/PropertyLookup.vue
@@ -20,7 +20,6 @@ import { MenuItem } from '@wmde/wikit-vue-components/dist/components/MenuItem';
 import Vue, { PropType } from 'vue';
 
 import { Lookup } from '@wmde/wikit-vue-components';
-import allowedDatatypes from '@/allowedDataTypes';
 import SearchResult from '@/data-access/SearchResult';
 
 export default Vue.extend( {
@@ -37,12 +36,12 @@ export default Vue.extend( {
 	watch: {
 		async search( newSearchString: string ): Promise<void> {
 			const searchResults = await this.$store.dispatch( 'searchProperties', newSearchString );
-			this.searchResults = searchResults.map( ( key: MenuItem & SearchResult ) => {
-				if ( !allowedDatatypes.includes( key.datatype ) ) {
-					key.tag = this.$i18n( 'query-builder-property-lookup-limited-support-tag' );
-				}
-				return key;
-			} );
+			this.searchResults = searchResults.map(
+				( item: MenuItem & SearchResult ) => {
+					item.tag = item.tag && this.$i18n( item.tag );
+					return item;
+				},
+			);
 		},
 	},
 	props: {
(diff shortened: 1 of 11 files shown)


src/components/PropertyLookup.vue
```diff
@@ -34,7 +36,13 @@ export default Vue.extend( {
 	},
 	watch: {
 		async search( newSearchString: string ): Promise<void> {
-			this.searchResults = await this.$store.dispatch( 'searchProperties', newSearchString );
+			const searchResults = await this.$store.dispatch( 'searchProperties', newSearchString );
+			this.searchResults = searchResults.map( ( key: MenuItem & SearchResult ) => {
```
micgro42: Can we move this into the action?

src/components/ValueTypeDropDown.vue
```diff
@@ -57,12 +72,7 @@ export default Vue.extend( {
 			width: 256px;
 			height: 32px;
 			margin-inline-start: $dimension-layout-xsmall;
-			margin-block-start: 23px;
-			border-color: #a2a9b1;
-			border-style: solid;
-			border-width: 1px;
-			border-radius: 2px;
-			appearance: auto;
+			margin-block-start: 21px;
```
bereket-WMDE: shouldn't all this style be removed ? 
AFAIK, the styles are already implemented in the DS. 
just like in the lookup component

src/components/ValueTypeDropDown.vue
```diff
@@ -4,20 +4,12 @@
 			class="querybuilder__dropdown-label">
 			Value Type
 		</label>
-		<select
+		<Dropdown
 			class="querybuilder__dropdown-select"
-			v-model="selected"
+			v-model="selectedItem"
```
micgro42: No, we can make all that much simpler now. We basically just listen to the `input` event and re-emit a cleaned version that contains only the value we care about. Conversely, we can set the `value` prop here directly.

Then we can also drop all the `watch` methods and the entire local state. We would then only have something like two computed methods and an event listener left

src/components/ValueTypeDropDown.vue
```diff
@@ -26,12 +18,19 @@
...
 	name: 'ValueTypeDropDown',
 	data() {
 		return {
-			selected: PropertyValueRelation.Matching,
-			optionItems: PropertyValueRelation,
+			selectedItem: {
+				label: PropertyValueRelation.Matching,
+				description: '',
+			},
+			optionItems: Object.values( PropertyValueRelation ).map( ( value: PropertyValueRelation ) => {
```
micgro42: Those option items should have translations. They could have the form of
```
{ 
  value: PropertyValueRelation.Matching,
  label: i18n('query-builder-property-lookup-' + PropertyValueRelation.Matching ),
}
```
 or something like that

micgro42 (changes requested):

src/components/ValueTypeDropDown.vue
Ladsgroup: I removed most of them, but If I remove the rest, it'll look like this:
image

src/components/PropertyLookup.vue
Ladsgroup: It needs i18n and that's not accessible in outside of the vue object. I can use it but it'll be pretty messy.

src/components/ValueTypeDropDown.vue
Ladsgroup: I wrestled with it for hours before making this PR, I admit I don't know vue well enough to do what you mean and make it work. Can you try it?

src/components/ValueTypeDropDown.vue
bereket-WMDE: hmm... @micgro42 could you verify that we don't have any style defined for the lookup component in the QB ?
if so, should it have been the same case with the dropdown ?
thanks

src/components/ValueTypeDropDown.vue
Ladsgroup: I noticed the problem way before the upgrade, I feel this is out of scope of this PR, we should fix it soon but separately

src/components/PropertyLookup.vue
micgro42: Mh, I'll have a look. Having the i18n here, but the assigning the tag in the action should work and already make things a lot nicer

src/components/ValueTypeDropDown.vue
micgro42: Will do!

src/components/ValueTypeDropDown.vue
micgro42: I think that is fair. Could you create a ticket in the current iteration board for this while I look at the other issues?

src/components/ValueTypeDropDown.vue
micgro42: will have a look

micgro42: I moved the responsibility of arranging the child components to the QueryCondition component

micgro42 (approved):

PR #96, opened 2020-12-04
Fix changes in query in state

Ladsgroup:
Without it, the run query button just stops working since validator is
not getting the right FormValues object

I think this was an implicit rebase issue that was overlooked.

For later, we should rework the validation to handle multiple condition
rows.

### DIFF
--- a/src/components/QueryBuilder.vue
+++ b/src/components/QueryBuilder.vue
@@ -37,6 +37,7 @@ import QueryResult from '@/components/QueryResult.vue';
 import buildQuery from '@/sparql/buildQuery';
 import Validator from '@/form/Validator';
 import AddCondition from '@/components/AddCondition.vue';
+import { Condition } from '@/sparql/QueryRepresentation';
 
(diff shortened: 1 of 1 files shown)


src/components/QueryBuilder.vue
```diff
@@ -55,12 +56,14 @@ export default ( Vue as VueConstructor<Vue & { $refs: { condition: InstanceType<
...
-			};
-			const validator = new Validator( formValues );
+			const formValues = this.$store.getters.query.conditions.map( ( condition: Condition ) => {
+				return {
+					property: condition.propertyId,
+					value: condition.value,
+					propertyValueRelation: condition.propertyValueRelation,
+				};
+			} );
+			const validator = new Validator( formValues[ 0 ] );
```
micgro42, line 66: Ok for now, but we should make sure that we rework this as soon as we actually have multiple queries.

micgro42 (approved):

PR #97, opened 2020-12-14
Translate value type relation

guergana:
+ first commit : added translations and descriptions
+ second commit: implemented translation

### DIFF
--- a/public/i18n/en.json
+++ b/public/i18n/en.json
@@ -14,5 +14,8 @@
   "query-builder-result-error-incomplete-form": "One or more fields are empty. Please complete the query or select a fitting field type.",
   "query-builder-result-error-missing-property": "Please select a property",
   "query-builder-result-error-missing-value": "Please enter a value",
-  "query-builder-add-condition": "Add condition"
+  "query-builder-add-condition": "Add condition",
+  "query-builder-value-type-relation-dropdown-matching": "matching",
+  "query-builder-value-type-relation-dropdown-without": "without",
+  "query-builder-value-type-relation-dropdown-regardless-of-value": "regardless of value"
 }
--- a/public/i18n/qqq.json
+++ b/public/i18n/qqq.json
@@ -19,5 +19,8 @@
 	"query-builder-result-error-incomplete-form": "Error message when the form is incomplete",
 	"query-builder-result-error-missing-property": "Error when property is missing",
 	"query-builder-result-error-missing-value": "Error when value is missing",
-	"query-builder-add-condition": "Text on the button that adds a condition"
+	"query-builder-add-condition": "Text on the button that adds a condition",
+	"query-builder-value-type-relation-dropdown-matching": "Text on dropdown option to have the Query Builder select items that match a value with the given property. See also {{msg-mw|query-builder-value-type-relation-dropdown-without}} and {{msg-mw|query-builder-value-type-relation-dropdown-regardless-of-value}}",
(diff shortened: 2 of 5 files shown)


public/i18n/en.json
```diff
@@ -14,5 +14,8 @@
   "query-builder-result-error-incomplete-form": "One or more fields are empty. Please complete the query or select a fitting field type.",
   "query-builder-result-error-missing-property": "Please select a property",
   "query-builder-result-error-missing-value": "Please enter a value",
-  "query-builder-add-condition": "Add condition"
+  "query-builder-add-condition": "Add condition",
+  "query-builder-valuetyperelation-dropdown-matching": "matching",
```
Ladsgroup: the "valuetyperelation" is a bit hard to read, can it have a - too plz?

src/components/ValueTypeDropDown.vue
```diff
@@ -42,7 +42,7 @@ export default Vue.extend( {
 		optionItems(): PropertyValueRelationMenuItem[] {
 			return Object.values( PropertyValueRelation ).map( ( value: PropertyValueRelation ) => {
 				return {
-					label: value, // FIXME: replace with i18n -> T269453
+					label: this.$i18n( `query-builder-valuetyperelation-dropdown-${value}` ),
```
Ladsgroup: please add all possible i18n keys here as comment, this would help in later finding orphan messages and finding out where an i18n message is used. 
See https://phabricator.wikimedia.org/T224429 and specially https://phabricator.wikimedia.org/T235502 for more context.

Ladsgroup (changes requested):
Thanks for doing it.

public/i18n/qqq.json
```diff
@@ -19,5 +19,8 @@
 	"query-builder-result-error-incomplete-form": "Error message when the form is incomplete",
 	"query-builder-result-error-missing-property": "Error when property is missing",
 	"query-builder-result-error-missing-value": "Error when value is missing",
-	"query-builder-add-condition": "Text on the button that adds a condition"
+	"query-builder-add-condition": "Text on the button that adds a condition",
+	"query-builder-valuetyperelation-dropdown-matching": "Text on dropdown option: matching",
+	"query-builder-valuetyperelation-dropdown-without": "Text on dropdown option: without",
```
micgro42: These could maybe be made more descriptive from the translator's point of view. We could extend them to say something like `Text on dropdown option to have the Query Builder select items that have the given property, but not with the given value. See also {{msg-mw|query-builder-valuetyperelation-dropdown-matching}} and {{msg-mw|query-builder-valuetyperelation-dropdown-regardless-of-value}}` and similarly for the other two

src/data-model/PropertyValueRelation.ts
```diff
@@ -1,7 +1,7 @@
 enum PropertyValueRelation {
```
micgro42, line 1: Could you add a comment here that these are also used as part of message keys and so any changes made to this enum have to also be reflected in `en.json` and `qqq.json`

micgro42 (commented):
Looks pretty good already

src/components/ValueTypeDropDown.vue
guergana: good point. fixing!

src/data-model/PropertyValueRelation.ts
guergana: ok!

public/i18n/qqq.json
guergana: Done!

micgro42 (approved):
Thank you!

Ladsgroup (approved):

PR #99, opened 2020-12-14
Remove static text direction declaration, dir=auto

bereket-WMDE:
Added dynamictext direction computing. The result is either ltr or rtl

Bug: T269438


bereket-WMDE:
@Ladsgroup decided to go the easy route. but I'm happy if you've further suggestions

Ladsgroup (approved):
https://caniuse.com/?search=getComputedStyle says it has a good support

PR #100, opened 2020-12-15
Fixed issue with delayed loading of new results on first scroll.

bereket-WMDE:
the user had to scroll back up again in order to load mode results

Bug: T268508

@micgro42 can you review this when you've time?
this is a fix for the loading of more results on the lookup, when a user scrolls for down.
previously, in order to load more results, the user had to scroll back up and down again.

### DIFF
--- a/src/components/PropertyLookup.vue
+++ b/src/components/PropertyLookup.vue
@@ -34,29 +34,33 @@ export default Vue.extend( {
 			search: '',
 			searchResults: [],
 			topItemIndex: 1,
-			scrollpropertiesSearchResults: [],
-			searchOptions: {} as SearchOptions,
 		};
 	},
 	methods: {
 		handleScroll( event: number ): void {
-			if ( event <= this.topItemIndex ) {
-				this.searchOptions.search = this.search;
-				this.searchOptions.offset = this.topItemIndex;
+			if ( this.topItemIndex <= event ) {
+				this.topItemIndex += 12;
 
-				this.searchPropertiesOnScroll( this.searchOptions );
-				this.searchResults = this.searchResults.concat( this.scrollpropertiesSearchResults );
-				this.topItemIndex += 13;
+				const searchOptions: SearchOptions = {
+					search: this.search,
+					offset: this.topItemIndex,
+				};
+
+				this.searchPropertiesOnScroll( searchOptions );
 			}
 		},
 		async searchPropertiesOnScroll( options: SearchOptions ): Promise<void> {
-			this.scrollpropertiesSearchResults = await this.$store.dispatch( 'searchProperties', options );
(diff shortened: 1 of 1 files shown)


src/components/PropertyLookup.vue
```diff
@@ -40,13 +40,12 @@ export default Vue.extend( {
 	},
 	methods: {
 		handleScroll( event: number ): void {
-			if ( event <= this.topItemIndex ) {
+			if ( this.topItemIndex <= event ) {
+				this.topItemIndex += 13;
```
micgro42: Shouldn't that be 12?

src/components/PropertyLookup.vue
```diff
@@ -64,6 +63,11 @@ export default Vue.extend( {
 				},
 			);
 		},
+		scrollpropertiesSearchResults( newSearchResults: object[] ): void {
```
micgro42: You don't need this watch, you can just move this if-condition as it is into `searchPropertiesOnScroll` and remake `this.scrollpropertiesSearchResults = ...` into a local variable `const searchResults = ...` and it should just work.

bereket-WMDE: yes. that was how i implemented it at first. However after it went through the test verification, it wasn't working. you had to scroll up again to see the items loaded again. 
after doing some debugging, i realized that, for some reason, the new scrollpropertiesSearchResults is never updated. i guess because of the async nature of the call?

that is why i ended up adding a watch.

bereket-WMDE: good call.
updated

src/components/PropertyLookup.vue
```diff
@@ -34,23 +34,22 @@ export default Vue.extend( {
...
-			scrollpropertiesSearchResults: [],
 			searchOptions: {} as SearchOptions,
 		};
 	},
 	methods: {
 		handleScroll( event: number ): void {
-			if ( event <= this.topItemIndex ) {
+			if ( this.topItemIndex <= event ) {
+				this.topItemIndex += 12;
 				this.searchOptions.search = this.search;
```
micgro42: `searchOptions` should not be a field, but should be local variables. Both here and below in the watcher.
Currently, it remembers the search position across searches with different strings, but for every new search string it should start from the beginning again

micgro42 (changes requested):
Sorry, the original implementation has another bug that I just noticed and that we probably want to fix in this PR as well

src/components/PropertyLookup.vue
bereket-WMDE: thanks for looking into that and checking it.
i made the changes as requested

micgro42 (approved):
Thank you :)

PR #101, opened 2020-12-15
Push validation into the store

micgro42:
Handling the validation in the components is becoming too complex with
multiple rows. Pushing it into the store separates the business logic
from the display logic.

### DIFF
--- a/src/store/actions.ts
+++ b/src/store/actions.ts
@@ -1,10 +1,11 @@
 import allowedDatatypes from '@/allowedDataTypes';
+import FormValues from '@/form/FormValues';
+import Validator from '@/form/Validator';
 import { MenuItem } from '@wmde/wikit-vue-components/dist/components/MenuItem';
 import { ActionContext } from 'vuex';
-import RootState from './RootState';
+import RootState, { ConditionRow } from './RootState';
 import SearchResult from '@/data-access/SearchResult';
 import Error from '@/data-model/Error';
-import Property from '@/data-model/Property';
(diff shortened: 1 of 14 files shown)


micgro42:
The `setProperty` mutation needs more thought :/

Ladsgroup (approved):
I'm slightly meh on the side that validation is being done asynchronously (by dispatching to store) but since the runQuery waits for the result, I have no complaints :D 

Thanks for this massive PR, lots of work.

src/store/actions.ts
```diff
@@ -31,8 +32,20 @@ export default ( searchEntityRepository: SearchEntityRepository, metricsCollecto
...
 	updateProperty( context: ActionContext<RootState, RootState>,
-		payload: { property: Property; conditionIndex: number } ): void {
+		payload: { property: { label: string; id: string; datatype: string }; conditionIndex: number } ): void {
+
 		context.commit( 'setProperty', payload );
+		if ( payload.property && !allowedDatatypes.includes( payload.property.datatype ) ) {
+			context.dispatch( 'setConditionAsLimitedSupport', payload.conditionIndex );
+		} else {
+			context.commit(
+				'setFieldErrors',
```
jakobw, line 42: For bonus points you could create a separate `clearFieldErrors` mutation instead of using `setFieldErrors` with `null`. I tripped over this line a tiny bit while reading through it.

I didn't manage to review this PR in detail but what I saw looked good! :)

micgro42: Done in #106 :)

PR #102, opened 2020-12-15
Move unsupported data types out of the allowed list

Ladsgroup:
They use a different RdfBuilder meaning the values wouldn't be matched.
We should be able to have better support later

Bug: T269439


guergana (approved):
Seems fair! LGTM.

PR #103, opened 2020-12-16
added methods searchEntities and searchItemValues

guergana:
this method is called by public searchProperties and searchItems methods:
adjust the repository to also allow querying for items
Add an action to use the repository to query for items
Updated searchEntityRepo interfase
Updated tests that use createActions
Added tests specific to the method searchItemValues

### DIFF
--- a/src/store/actions.ts
+++ b/src/store/actions.ts
@@ -25,6 +25,15 @@ export default ( searchEntityRepository: SearchEntityRepository, metricsCollecto
 			return searchResult;
 		} );
 	},
+	async searchItemValues(
+		_context: ActionContext<RootState, RootState>,
+		options: SearchOptions ): Promise<SearchResult[]> {
+		// check for empty
+		const searchResults = await searchEntityRepository.searchItemValues( options.search, 12, options.offset );
+		return searchResults.map( ( searchResult: MenuItem & SearchResult ) => {
+			return searchResult;
+		} );
+	},
 	updateValue(
 		context: ActionContext<RootState, RootState>,
 		payload: { value: string; conditionIndex: number } ): void {
--- a/src/store/actions.ts
+++ b/src/store/actions.ts
@@ -29,10 +29,7 @@ export default ( searchEntityRepository: SearchEntityRepository, metricsCollecto
 		_context: ActionContext<RootState, RootState>,
 		options: SearchOptions ): Promise<SearchResult[]> {
 		// check for empty
-		const searchResults = await searchEntityRepository.searchItemValues( options.search, 12, options.offset );
-		return searchResults.map( ( searchResult: MenuItem & SearchResult ) => {
-			return searchResult;
-		} );
+		return await searchEntityRepository.searchItemValues( options.search, 12, options.offset );
 	},
 	updateValue(
 		context: ActionContext<RootState, RootState>,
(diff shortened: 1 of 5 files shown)


src/store/actions.ts
```diff
@@ -25,6 +25,15 @@ export default ( searchEntityRepository: SearchEntityRepository, metricsCollecto
 			return searchResult;
 		} );
 	},
+	async searchItemValues(
+		_context: ActionContext<RootState, RootState>,
+		options: SearchOptions ): Promise<SearchResult[]> {
+		// check for empty
+		const searchResults = await searchEntityRepository.searchItemValues( options.search, 12, options.offset );
```
guergana: i don't know where this 12 comes from, i took it from `searchProperties`, does it make sense here?
I guess it is because it the visible amount in two full height scrolls. (please confirm)

micgro42: I think UX decided that we always want to load results in chunks of 12. In a follow-up we should probably introduce a constant for that.

src/store/actions.ts
```diff
@@ -25,6 +25,15 @@ export default ( searchEntityRepository: SearchEntityRepository, metricsCollecto
...
+		return searchResults.map( ( searchResult: MenuItem & SearchResult ) => {
```
micgro42: This map does nothing, we don't need it here.

tests/unit/data-access/FetchSearchEntityRepository.spec.ts
```diff
@@ -89,16 +159,52 @@ describe( 'FetchSearchEntityRepository', () => {
...
+			statusText: 'Server Error',
+		} ) );
+
+		const expectedError = new TechnicalProblem( '500: Server Error' );
 
-		const expectedError = new TechnicalProblem( 'Network error' );
+		expect( repo.searchItemValues( 'John Smith' ) ).rejects.toThrow( expectedError );
+	} );
+
+	it( 'throws an error if there is a server side problem fetching a property', () => {
```
micgro42: No, this test should be about network errors, not server errors

tests/unit/store/actions.spec.ts
```diff
@@ -80,6 +86,32 @@ describe( 'actions', () => {
...
+		it( 'calls the repo and resolves with the result', async () => {
+			const expectedResult = [ { label: 'potato', id: 'Q666', datatype: 'string' } ];
+			const searchProperties = jest.fn().mockResolvedValue(
+				JSON.parse( JSON.stringify( expectedResult ) ),
+			);
+			const searchItemValues = jest.fn().mockResolvedValue(
+				JSON.parse( JSON.stringify( expectedResult ) ),
+			);
+			const actions = createActions(
+				{ searchProperties, searchItemValues },
```
micgro42: You can replace the `searchProperties` argument just with a generic function stub since it is never called anyway. Similarly in the other modified tests in this file.

```suggestion
				{ jest.fn(), searchItemValues },
```

micgro42 (changes requested):
Nice! There are just some minor things to adjust

src/store/actions.ts
guergana: That sounds like a very good idea!

src/store/actions.ts
guergana: You couldn't be more right!

tests/unit/data-access/FetchSearchEntityRepository.spec.ts
guergana: I deleted that one by accident it seems. Also duplicated one. It's a copy and paste mistake. Will fix.

micgro42 (approved):
Thank you :blush:

PR #106, opened 2020-12-18
Add clearFieldErrors mutation to be more purposeful

micgro42:
This way we don't have to misuse the setFieldErrors mutation for clearing them.


jakobw (approved):
:+1:

PR #107, opened 2020-12-18
Build triple objects for wikibase-item values

micgro42:
Decided to not extract the dispatcher for now and to wait till we start dealing with unit values.

Bug: T270255


guergana (approved):
looks good!

PR #109, opened 2020-12-18
Don't search for an empty string

micgro42:
This error happened when the user typed something into the Lookup and then deleted it all again.


guergana (approved):
yei, it works now!

PR #110, opened 2020-12-18
Extract magic number

micgro42:
We should not have random numbers "12" floating around in our code.

### DIFF
--- a/src/components/PropertyLookup.vue
+++ b/src/components/PropertyLookup.vue
@@ -24,6 +24,8 @@ import { Lookup } from '@wmde/wikit-vue-components';
 import SearchResult from '@/data-access/SearchResult';
 import SearchOptions from '@/data-access/SearchOptions';
 
+const NUMBER_OF_SEARCH_RESULTS = 12;
+
 export default Vue.extend( {
 	name: 'PropertyLookup',
 	components: {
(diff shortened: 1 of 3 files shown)


src/Constants.ts
```diff
@@ -0,0 +1,6 @@
+/**
+ * This is a collection of values that are constant for this application,
+ * but typically would be configurable
+ */
+
+export const NUMBER_OF_SEARCH_RESULTS = 12;
```
Ladsgroup: I don't think this should be in src/Constants.ts, it'll make the file slowly incohesive. (Principle of good design: components should be loosely coupled and highly cohesive). This will make the file slowly become a kitchen sink of different values from everywhere, that's why util files are also bad. I think this should go to fetch data for example.

Ladsgroup (changes requested):

src/Constants.ts
micgro42: This kind of value should be configured, i.e. go into .env.production

But since we don't actually want to configure it, I'm not sure if that is worth it?

Ladsgroup: I'm not 100% sure if this should be configurable. There is a whole school against making things configurable because it gives the software more freedom meaning it'll be less predictable and testable (you'll don't know if it's set 100, what would happen, it'll also implies people know what this configuration means.). According to a DevOps survey in google, 50% of their outages was due to misconfiguration (we had our fair share of incidents because of it). So I don't think this should be configurable.

This should be in the place that uses it the most, like the class that fetches it or part of interface of search result.

micgro42: :thinking:

micgro42: I think I know where to put it!

Ladsgroup (approved):
Thanks <3

PR #111, opened 2020-12-18
Extract EntityLookup and add ItemValueLookup

micgro42:
ItemValueLookup and PropertyLookup share a lot of logic, yet there are
subtle differences, both in logic and in semantics.

This is passing down a method as a prop. While that might be more
unusual in vuejs than in react, it seems that having the parent maintain
the state only for it to be passed down to the child and having the
child ask for more items on scroll would seem to be much more convoluted
and clumsy.

### TODO:
- [ ] agree on the basic approach
- [x] tests
- [x] messages

Bug: T270256

### DIFF
--- a/src/components/EntityLookup.vue
+++ b/src/components/EntityLookup.vue
@@ -1,0 +1,106 @@
+<template>
+	<Lookup
+		:value="value"
+		@input="$emit( 'input', $event )"
+		:error="error"
+		:menu-items="searchResults"
+		:search-input.sync="search"
+		:placeholder="placeholder"
+		:label="label"
+		v-on:scroll="handleScroll"
+	>
+		<template
+			v-slot:no-results
+		>{{ noMatchFoundMessage }}</template>
+	</Lookup>
+</template>
+
+<script lang="ts">
+import { MenuItem } from '@wmde/wikit-vue-components/dist/components/MenuItem';
+import Vue, { PropType } from 'vue';
+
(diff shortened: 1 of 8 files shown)


Ladsgroup:
I'm not 100% we should do this. The amount of duplicated code is small and if it grows, it should be moved the pure ts/js instead of being in Vue. Some of the code you put here can be moved to pure ts/js and reuse that in the components.

micgro42:
Rebased it on top of master

micgro42:
> I'm not 100% we should do this. The amount of duplicated code is small and if it grows, it should be moved the pure ts/js instead of being in Vue. Some of the code you put here can be moved to pure ts/js and reuse that in the components.

Yeah, at some point we might want to share code with a "Unit Lookup" :thinking:

src/components/EntityLookup.vue
```diff
@@ -0,0 +1,106 @@
...
+			}
+			const searchOptions: SearchOptions = {
+				search: newSearchString,
+				limit: NUMBER_OF_SEARCH_RESULTS,
+			};
+			this.searchResults = await this.searchEntities( searchOptions );
+		},
+	},
+	props: {
+		searchForMenuItems: {
```
Ladsgroup, line 80: Passing it as a prop. Nice idea!

Ladsgroup (approved):
Thanks!

PR #114, opened 2020-12-21
Fixed wrong display of "limited support" tag on new results of

bereket-WMDE:
the lookup menu upon scroll

Bug: T270549

### DIFF
--- a/src/components/PropertyLookup.vue
+++ b/src/components/PropertyLookup.vue
@@ -34,7 +34,7 @@ export default Vue.extend( {
 	data() {
 		return {
 			search: '',
-			searchResults: [],
+			searchResults: [] as SearchResult[],
 			topItemIndex: 1,
 		};
 	},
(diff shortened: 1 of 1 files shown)


src/components/PropertyLookup.vue
```diff
@@ -53,7 +53,14 @@ export default Vue.extend( {
 			}
 		},
 		async searchPropertiesOnScroll( options: SearchOptions ): Promise<void> {
-			const searchResults = await this.$store.dispatch( 'searchProperties', options );
+			let searchResults = await this.$store.dispatch( 'searchProperties', options );
+			searchResults = searchResults.map(
```
guergana: this has the exact same content as [the watch](https://github.com/wmde/query-builder/blob/51673294a6fa2c0893c73a3752748ddf0de11692/src/components/PropertyLookup.vue#L71). It is better if it's extracted into a function.

guergana (changes requested):

src/components/PropertyLookup.vue
bereket-WMDE: updated.

src/components/PropertyLookup.vue
```diff
@@ -34,7 +34,7 @@ export default Vue.extend( {
 	data() {
 		return {
 			search: '',
-			searchResults: [],
+			searchResults: [] as object[],
```
guergana: why cast type it to `object[]` instead of `SearchResult[]` ?

src/components/PropertyLookup.vue
```diff
@@ -53,9 +53,19 @@ export default Vue.extend( {
 			}
 		},
 		async searchPropertiesOnScroll( options: SearchOptions ): Promise<void> {
-			const searchResults = await this.$store.dispatch( 'searchProperties', options );
+			let searchResults = await this.$store.dispatch( 'searchProperties', options );
+			searchResults = this.setTagForSearchResults( searchResults );
+
 			this.searchResults = this.searchResults.concat( searchResults );
 		},
+		setTagForSearchResults( searchResults: SearchResult[] ): object[] {
```
guergana: same observation as before: `SearchResult[]` instead of object[] is a better type in this case.

guergana (changes requested):

bereket-WMDE:
@guergana good catch on the type casting comments!

guergana (approved):

PR #115, opened 2020-12-21
improve legibility of description text by limiting the line-width

guergana:


src/components/QueryBuilder.vue
```diff
@@ -122,6 +123,12 @@ a {
 	font-size: $font-size-style-description;
 	line-height: $font-line-height-style-description;
 	color: $font-color-base;
+	max-width: 672px;
```
guergana: Didn't find a token for this but happy to replace it if there is a plan to add one. Maybe we can merge and replace with token in the future?

micgro42: Yep, we can always replace it with a token later. But could you add a comment to it? Maybe something like:

```suggestion
	max-width: 672px; // TODO: replace with token
```

So that we stumble over it and remember to ask a designer about it

micgro42 (approved):

PR #116, opened 2020-12-22
implemented remove condition

guergana:
+ quiet button will be added when new version of wikit is released
+ added removeCondition action and mutation
+ added tests

### DIFF
--- a/src/components/QueryCondition.vue
+++ b/src/components/QueryCondition.vue
@@ -19,12 +19,19 @@
 			:placeholder="$i18n('query-builder-input-value-placeholder')"
 			:disabled="isTextInputDisabled()"
 		/>
+		<Button
+			class="query-condition__remove"
+			type="primaryDestructive"
+			@click.native="removeCondition"
+			:disabled="!canDelete"
+		> Trash
+		</Button>
 	</div>
 </template>
 
 <script lang="ts">
 import Vue from 'vue';
-import { TextInput } from '@wmde/wikit-vue-components';
+import { TextInput, Button } from '@wmde/wikit-vue-components';
 
 import PropertyLookup from '@/components/PropertyLookup.vue';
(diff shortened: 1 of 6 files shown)


src/components/QueryCondition.vue
```diff
@@ -19,12 +19,19 @@
 			:placeholder="$i18n('query-builder-input-value-placeholder')"
 			:disabled="isTextInputDisabled()"
 		/>
+		<Button
```
micgro42, line 22: This should probably be its own simple subcomponent as well, similar to ValueTypeDropdown and PropertyLookup. The idea is to try to maintain "one level of abstraction" within the template as well.

But that could also be done in a follow-up pull request

tests/unit/store/mutations.spec.ts
```diff
@@ -97,6 +97,37 @@ describe( 'mutations', () => {
...
 
+	it( 'removeCondition', () => {
+		const keptRow = {
+			valueData: { value: 'foo', valueError: null },
+			propertyData: { id: 'P123', label: 'abc', datatype: 'string', propertyError: null },
+			propertyValueRelationData: { value: PropertyValueRelation.Matching },
+			conditionId: '0.123',
+		};
+		const state: RootState = {
+			conditionRows: [ {
```
micgro42: You can replace this first object with the `keptRow` variable from above :)

micgro42 (commented):
Thank you for taking this on! We should extract the component, but that should be relatively

src/components/QueryCondition.vue
guergana: Let's do it in a followup. :)

micgro42: :+1:

micgro42 (approved):

PR #117, opened 2020-12-23
modified the store to be able to change the limit

guergana:
added a getter (to get the default in case it is not selected)
add an action
added a mutation

### DIFF
--- a/src/store/getters.ts
+++ b/src/store/getters.ts
@@ -55,6 +55,9 @@ export default {
 			return rootState.conditionRows[ conditionIndex ].propertyValueRelationData.value;
 		};
 	},
+	limit( rootState: RootState ): ( number | undefined ) {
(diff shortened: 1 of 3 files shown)


src/store/getters.ts
```diff
@@ -55,6 +55,9 @@ export default {
 			return rootState.conditionRows[ conditionIndex ].propertyValueRelationData.value;
 		};
 	},
+	limit( rootState: RootState ): ( number | undefined ) {
```
micgro42, line 58: I feel like I see those parentheses around the types union for the first time, but it seems we used them before, so it's fine :)

micgro42 (approved):

src/store/getters.ts
guergana: It's the holidays, everything seems so new. :D

PR #118, opened 2020-12-23
Limit UI

guergana:
number validation to be done in another PR. This is already too big.

Commits 2-4 were part of [PR 119](https://github.com/wmde/query-builder/pull/119) and have been merged into this PR and were already reviewed by @micgro42

### DIFF
--- a/public/i18n/en.json
+++ b/public/i18n/en.json
@@ -19,5 +19,6 @@
   "query-builder-value-type-relation-dropdown-matching": "matching",
   "query-builder-value-type-relation-dropdown-without": "without",
   "query-builder-value-type-relation-dropdown-regardless-of-value": "regardless of value",
-  "query-builder-delete-condition": "Delete condition"
+  "query-builder-delete-condition": "Delete condition",
+  "query-builder-limit-number-results-description": "Limit the number of results to"
 }
--- a/src/components/Limit.vue
+++ b/src/components/Limit.vue
@@ -1,0 +1,74 @@
+<template>
+	<div class="querybuilder__limit">
+		<input
+			class="querybuilder__limit-checkbox"
+			type="checkbox"
+			id="limit"
+			v-model="checked"
+		/>
+		<label for="limit">
+			<span
+				v-i18n="{msg: 'query-builder-limit-number-results-description'}" />
+		</label>
+		<TextInput
+			class="querybuilder__limit-input"
+			v-model="textInputValue"
+			:disabled="!checked"
+			:label="$i18n('query-builder-limit-number-results-description')"
+		/> <!-- TODO: decide how the missing label should be handled -->
+	</div>
+</template>
+
+<script lang="ts">
+import Vue from 'vue';
+import { TextInput } from '@wmde/wikit-vue-components';
+
+export default Vue.extend( {
+	name: 'Limit',
+	components: {
+		TextInput,
+	},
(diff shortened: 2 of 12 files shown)


public/i18n/en.json
```diff
@@ -17,5 +17,6 @@
   "query-builder-add-condition": "Add condition",
   "query-builder-value-type-relation-dropdown-matching": "matching",
   "query-builder-value-type-relation-dropdown-without": "without",
-  "query-builder-value-type-relation-dropdown-regardless-of-value": "regardless of value"
+  "query-builder-value-type-relation-dropdown-regardless-of-value": "regardless of value",
+  "query-builder-limit-number-results-description": "Limit the number of results to"
```
micgro42, line 23: I think the copy doesn't really belong in this PR. Mainly, because I'm not sure that it is already final and so we shouldn't make translators work on things that might be outdated by tomorrow. Let's include that copy as literal text in the component for now

micgro42 (changes requested):

public/i18n/en.json
guergana: It is not clear in the tickets that this is not the final copy. There is an older version somewhere but this is final as far as I am concerned. Did I miss some discussion about this?

micgro42: Yes, this part of the ticket wasn't ready to be worked on in the first place. I think there were still some UX/design questions open about it. At the very least, we need a good idea for how to split the meaning of that single label between input components.

guergana: We can discuss it tomorrow in the meeting and leave it here for now maybe

guergana: It has been decided that this is the final text for now.

src/components/Limit.vue
```diff
@@ -14,8 +14,8 @@
 			class="querybuilder__limit-input"
 			v-model="textInputValue"
 			:disabled="!checked"
-			:label="$i18n('query-builder-limit-number-results-description')"
-		/> <!-- TODO: decide how the missing label should be handled -->
+			label="Maximum number of results"
```
guergana, line 17: at some point we need to also translate this. Also decide the final text

src/components/Limit.vue
```diff
@@ -0,0 +1,87 @@
+<template>
+	<div class="querybuilder__limit">
```
micgro42, line 2: this classname isn't exactly ideal BEM, but we have this problem in other places as well. We'll need to do a general sweep of class names to align all of them. But not in this PR.

src/store/RootState.ts
```diff
@@ -5,6 +5,7 @@ export default interface RootState {
 	conditionRows: ConditionRow[];
 	errors: Error[];
 	limit?: number;
```
micgro42: It makes sense to make that field non-optional right away. We have to touch every test where this is relevant anyway to add the `useLimit` field, so we can add this as well.

micgro42 (changes requested):
Only a small thing to change: making the `limit` field in the store not optional. This should only require another line to some tests that you already touched

After that it can be merged :)

src/components/Limit.vue
guergana: oh, i was taking other components as guide. O: yes, makes sense to change them all at once if this is not the best way

micgro42 (approved):

PR #119, opened 2020-12-23
Limit component

guergana:
added limit component. 

Still missing some tests. 

The validation of the number in the field needs to be discussed as well, also if `limit?` is required or not and if it can be required conditionally when `useLimit == true`

### DIFF
--- a/src/components/Limit.vue
+++ b/src/components/Limit.vue
@@ -1,0 +1,70 @@
+<template>
+	<div class="querybuilder__limit">
+		<input
+			class="querybuilder__limit-checkbox"
+			type="checkbox"
+			id="limit"
+			v-model="checked"
+		/>
+		<label for="limit">
+			<span
+				v-i18n="{msg: 'query-builder-limit-number-results-description'}" />
+		</label>
+		<TextInput
+			class="querybuilder__limit-input"
+			v-model="textInputValue"
+			:disabled="!checked"
+			:label="$i18n('query-builder-limit-number-results-description')"
+		/>
+	</div>
+</template>
+
+<script lang="ts">
+import Vue from 'vue';
+import { TextInput } from '@wmde/wikit-vue-components';
+
+export default Vue.extend( {
+	name: 'Limit',
+	components: {
+		TextInput,
+	},
+	data() {
+		return {
+			checked: true,
+		};
+	},
+	computed: {
+		textInputValue: {
+			// TODO: change after deciding how to validate numbers for TextInput
+			get(): ( string | undefined ) {
+				return this.$store.getters.limit?.toString();
+			},
+			set( value: string ): void {
+				this.$store.dispatch( 'setLimit', parseInt( value ) );
+			},
+		},
+	},
+} );
+</script>
+
+<style scoped lang="scss">
+
+.querybuilder__limit {
+	display: flex;
+	align-items: flex-end;
+	color: $font-color-base;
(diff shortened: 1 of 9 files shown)


src/sparql/QueryRepresentation.ts
```diff
@@ -9,4 +9,5 @@ export type Condition = {
 export default interface QueryRepresentation {
 	conditions: Condition[];
 	limit?: number;
+	useLimit: boolean;
```
micgro42: We might need the `useLimit` flag in the store (not sure yet), but here in the `QueryRepresentation`, we can just set `limit` to `null` or omit it, if we do not want to have a limit there.

guergana: I added it because there is a default of 100. In that case useLimit is true and taken into account since the start.

micgro42: But isn't that correct? My Idea would be something like the following:

<table>
<tr>
	<th>State
	<th>`QueryRepresentation`
<tr>
	<td>`limit = 100; useLimit = true` (the default)
	<td>`limit = 100`
<tr>
	<td>`limit = 100; useLimit = false` (checkbox unchecked)
	<td>`limit = null`
<tr>
	<td>`limit = 30; useLimit = true` (checkbox checked and value changed)
	<td>`limit = 30`
<tr>
	<td>`limit = 30; useLimit = false` (checkbox unchecked and value changed)
	<td>`limit = null`
</table>

guergana: ah, now I understand. Yes. That's true. Thanks for explaining.

src/components/Limit.vue
```diff
@@ -0,0 +1,70 @@
+<template>
+	<div class="querybuilder__limit">
+		<input
+			class="querybuilder__limit-checkbox"
+			type="checkbox"
+			name="limit"
+			v-model="checked"
+		/>
+		<label for="limit">
```
micgro42, line 9: the `for` attribute refers to an id, not a name, see https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label#attr-for

micgro42 (changes requested):
It is a bit strange to add the action and mutation in the same PR as the component, but then not use it there. But since that is done in the following PR, it is not worth splitting this PR up

src/components/Limit.vue
guergana: This will be merged in the other PR eventually. This is a mistake yes..

guergana: This will be replaced, as the label, and even the component are going to be replaced. Will update it but I dont think it's worth spending much time on this part, tomorrow it will be probably decided how to handle this.

micgro42: If the idea is to merge #120 into this PR first, then I'm ok with it :)

guergana: Oh, no. :sweat_smile:  The idea was to merge this one into [PR 118](https://github.com/wmde/query-builder/pull/118). I've updated this PR

micgro42 (approved):
I think this is now ok by itself, though there is still the question about using messages vs string literals for the copy for now, which is discussed in #118

guergana:
> I think this is now ok by itself, though there is still the question about using messages vs string literals for the copy for now, which is discussed in #118

Let's ask the designers and then take a decision :)

PR #120, opened 2020-12-28
Limit component

guergana:
Includes several commits. PR 119 was merged here but then did some rearranging, this is why it is not visible in the commit history

### DIFF
--- a/tests/unit/components/Limit.spec.ts
+++ b/tests/unit/components/Limit.spec.ts
@@ -1,0 +1,84 @@
+import Vue from 'vue';
+import Limit from '@/components/Limit.vue';
+import { TextInput } from '@wmde/wikit-vue-components';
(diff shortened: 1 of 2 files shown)


src/components/Limit.vue
```diff
@@ -3,7 +3,7 @@
 		<input
 			class="querybuilder__limit-checkbox"
 			type="checkbox"
-			name="limit"
+			id="limit"
```
guergana: This id also would go away once it's decided how the label will be handled. This will be replaced with the checkbox component anyway

micgro42: This change to that attribute should be in the previous PR #119

guergana:
I will squash this into 118, it doesn't make sense to have in separate PRs now.

PR #122, opened 2021-01-04
Extract DeleteConditionButton for better encapsulation

micgro42:
This should contribute towards having all the template in the QueryCondition component be on a single level of abstraction.

Bug: [T271093](https://phabricator.wikimedia.org/T271093)


guergana (approved):
Looks great!

PR #123, opened 2021-01-04
Dispatch Input based in property datatype

micgro42:
(To be merged into master after #142)

Todo:
* [x] agree on the basic approach
* [x] tests

Bug: T270257

### DIFF
--- a/src/components/ValueInput.vue
+++ b/src/components/ValueInput.vue
@@ -1,0 +1,51 @@
+<template>
+	<component
+		:is="dispatchComponent"
+		:value="value"
+		@input="$emit('input', $event)"
+		:disabled="disabled"
+		:error="error"
+	/>
+</template>
+
+<script lang="ts">
+import ItemValueLookup from '@/components/ItemValueLookup.vue';
+import StringValueInput from '@/components/StringValueInput.vue';
+import Vue from 'vue';
+
+export default Vue.extend( {
+	name: 'ValueInput',
+	components: {
+		StringValueInput,
+		ItemValueLookup,
+	},
+	computed: {
+		dispatchComponent(): string {
+			switch ( this.datatype ) {
+				case 'wikibase-item':
+					return 'ItemValueLookup';
+				case 'string':
+				case 'external-id':
+				default:
+					return 'StringValueInput';
+			}
+		},
+	},
+	props: {
+		datatype: {
+			required: true,
+		},
+		value: {
+			default: null,
+		},
+		error: {
+			type: Object,
+			default: null,
+		},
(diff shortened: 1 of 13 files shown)


src/components/ValueInput.vue
```diff
@@ -0,0 +1,48 @@
...
+	/>
+	<TextInput
+		v-else
+		:value="value"
+		:label="$i18n('query-builder-input-value-label')"
+		@input="$emit('input', $event)"
+		:error="error && { message: $i18n(error.message), type: error.type }"
+		:placeholder="$i18n('query-builder-input-value-placeholder')"
+		:disabled="disabled"
+	/>
```
jakobw: If value input components share the same interface (which they probably should) this whole template (or the whole component?) could turn into something like this:

```suggestion
	<component
		:is="{
			string: 'TextInput',
			'external-id': 'TextInput',
			'wikibase-item': 'ItemValueLookup'
		}[datatype]"
		:value="value"
		:label="$i18n('query-builder-input-value-label')"
		@input="$emit('input', $event)"
		:error="error && { message: $i18n(error.message), type: error.type }"
		:placeholder="$i18n('query-builder-input-value-placeholder')"
		:disabled="disabled"
	/>
```
... but ideally without doing the dispatching in the template.

If that works out, all this component does is forwarding props. At that point it seems cleaner to put the dispatching code into a regular function, and move the differences we see in the templates (error handling, messages, ...) down into the value input components. Then this can become a [functional component](https://vuejs.org/v2/guide/render-function.html#Functional-Components) or be removed entirely if nothing is left.

jakobw (commented):
Looks good! I tried it out locally and it works nicely! :trophy: 

I left a suggestion in there for the value input dispatching code. I don't think this necessarily has to happen in this PR but I'm curious to hear thoughts.

src/components/ValueInput.vue
micgro42: I really like the looks of this, thanks! I had looked at `<component is=""` previously and it didn't seem exactly right, but your approach is very convincing.

Then the first step for this would be to wrap the `TextInput` into its own component to align the interfaces, right?. That was on my Todo-list anyway, I will look into it.

jakobw: > Then the first step for this would be to wrap the TextInput into its own component to align the interfaces, right?

Yes, that sounds good!

jakobw (changes requested):
I think this looks super good! ValueInput.vue could use a test though, and it should be very nicely testable now.

jakobw (approved):
Woo! :fire:

PR #124, opened 2021-01-04
Change flex items alignment to start instead of end

Ladsgroup:
This way, if there's a validation message in the property input, the
rest of items wouldn't get pushed down.

This change actually makes the value type dropdown to be slightly
misaligned (by 0.25em) that's caused by the internal label in the
dropdown in wikit and needs to be fixed inside separately

### DIFF
--- a/src/components/QueryCondition.vue
+++ b/src/components/QueryCondition.vue
@@ -120,7 +120,7 @@ export default Vue.extend( {
 
 	.query-condition {
 		display: flex;
-		align-items: flex-end;
+		align-items: flex-start;
 		margin-block-start: $dimension-layout-medium;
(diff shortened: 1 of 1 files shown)


micgro42 (approved):
Better than what we had before 👍

We probably want to revisit the labels on the ValueTypeDropdown altogether. Having that extra label doesn't seem to make much sense.

We could either:
1: use the label property on the DS `Dropdown`, but hide it with the css that is currently in `ValueTypeDropdown`
or
2: use the `aria-label` directly (though we need to double-check that this _actually_ works with a screen reader)

Ladsgroup:
Went with the former: https://github.com/wmde/query-builder/pull/129

PR #125, opened 2021-01-05
Use flexbox to wrap the content vertically in property lookup

Ladsgroup:
This doesn't fix the issue 100% but moves towards it


guergana:
> This doesn't fix the issue 100% but moves towards it

:smile:

micgro42 (approved):
Not perfect and this might be wholly undone as soon as we start to use CSS grid, but since it already exists it looks nicer than not having it.

guergana:
This creates another problem. See: 

image

PR #127, opened 2021-01-06
Make datatype a proper ConditionRow property

micgro42:
This will allow us to make use of the datatype even when the property is unset and lets us act differently depending on whether the new and the old datatype are the same or not.

### DIFF
--- a/src/store/RootState.ts
+++ b/src/store/RootState.ts
@@ -10,6 +10,7 @@ export default interface RootState {
 
 export interface ConditionRow {
 	propertyData: PropertyData;
+	datatype: string|null;
 	valueData: {
 		value: string;
 		valueError: Error|null;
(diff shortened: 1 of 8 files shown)


Ladsgroup:
Can we make datatype part of the propertyData itself. I think it's a property of the property and searchresult already returns it as part of the propertyData

micgro42:
> Can we make datatype part of the propertyData itself. I think it's a property of the property and searchresult already returns it as part of the propertyData

The idea of this PR is to remove it from property data and make it a stand-alone thing. Yes, conceptually they are part of the property data, but the challenge is that our process is"select property -> set property to null -> select new property" and we need to remember the "old" datatype until we select the new property.

micgro42:
Replaced by #140, as @Ladsgroup was right and the datatype should indeed be part of the property.

PR #128, opened 2021-01-06
Explicitly unset property in store 

micgro42:
(To be merged into master after #140 and is followed by #141 )

We were already doing that implicitly, because the existing
updateProperty action is already being called with null as the property
without typescript reflecting that.

This now extracts that functionality into its own action and mutation
making the overall flow simpler.

There are no behavioral changes in this commit.

### DIFF
--- a/src/components/QueryCondition.vue
+++ b/src/components/QueryCondition.vue
@@ -85,7 +85,11 @@ export default Vue.extend( {
 			get(): SearchResult | null {
 				return this.$store.getters.property( this.conditionIndex );
 			},
-			set( selectedProperty: SearchResult ): void {
+			set( selectedProperty: SearchResult | null ): void {
+				if ( selectedProperty === null ) {
+					this.$store.dispatch( 'unsetProperty', this.conditionIndex );
+					return;
+				}
 				this.$store.dispatch(
 					'updateProperty',
 					{ property: selectedProperty, conditionIndex: this.conditionIndex },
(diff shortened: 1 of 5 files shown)


Ladsgroup (approved):

Ladsgroup (commented):
markPropertyAsUnset seems to have redundancy, you can simply say "unsetProperty"

micgro42:
> markPropertyAsUnset seems to have redundancy, you can simply say "unsetProperty"

I feel `unsetProperty` implies that the property data won't be available anymore, but it still is.

Ladsgroup:
> > markPropertyAsUnset seems to have redundancy, you can simply say "unsetProperty"
> 
> I feel `unsetProperty` implies that the property data won't be available anymore, but it still is.

I'm not a native speaker but I'm not sure if that's correct. Should we ask a native speaker?

micgro42:
> > > markPropertyAsUnset seems to have redundancy, you can simply say "unsetProperty"
> > 
> > 
> > I feel `unsetProperty` implies that the property data won't be available anymore, but it still is.
> 
> I'm not a native speaker but I'm not sure if that's correct. Should we ask a native speaker?

Sounds good to me. I think @addshore or @tarrow would be the obvious choice to ask for English language help?

micgro42:
The action and mutation have been renamed to `unsetProperty`

Ladsgroup (approved):

addshore:
> The action and mutation have been renamed to `unsetProperty`

Sounds good to me!

PR #129, opened 2021-01-06
Remove the extra label and put it in dropdown label for value type

Ladsgroup:
This way we avoid the extra 4 pixels caused by the hidden label


micgro42 (approved):
Looks good :+1:

PR #130, opened 2021-01-06
Use the trash icon

Ladsgroup:
Now we have the new wikit version, we can use it.

Bug: T268982

### DIFF
--- a/src/components/DeleteConditionButton.vue
+++ b/src/components/DeleteConditionButton.vue
@@ -1,16 +1,18 @@
 <template>
 	<Button
+		iconOnly
 		class="delete-condition-button"
-		type="destructive"
-		variant="primary"
+		type="neutral"
+		variant="quiet"
+		:aria-label="$i18n('query-builder-delete-condition')"
 		@click.native="$emit('click')"
 		:disabled="disabled"
-	> Trash
+	> <Icon type="trash" size="large" color="inherit" />
 	</Button>
 </template>
 
 <script lang="ts">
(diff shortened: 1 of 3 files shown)


src/components/DeleteConditionButton.vue
```diff
@@ -1,16 +1,18 @@
 <template>
 	<Button
+		iconOnly
 		class="delete-condition-button"
-		type="destructive"
-		variant="primary"
+		type="normal"
```
micgro42: This should be `type="neutral"` not normal, sorry!

src/components/DeleteConditionButton.vue
```diff
@@ -1,16 +1,18 @@
...
+		variant="quiet"
+		aria-label="Delete condition"
```
micgro42: Could this be an i18n message?

micgro42 (changes requested):

src/components/DeleteConditionButton.vue
guergana: +1

src/components/DeleteConditionButton.vue
Ladsgroup: Yeah, noticed it late :D

src/components/DeleteConditionButton.vue
Ladsgroup: Done

micgro42 (approved):

PR #131, opened 2021-01-07
Grid layout for condition row

micgro42:
This adds the placeholders for the planned toggle buttons and references
Dropdown in order to be able to take them into account for the layout.

The magic number for the $clearWidth comes from the magic number for the
intro text. Not ideal, but works for now. Maybe this should be replace
by some em value?

This doesn't work too nicely for very small screens, but it is still
usable.

Bug: [T269470](https://phabricator.wikimedia.org/T270179)

### DIFF
--- a/src/components/PropertyLookup.vue
+++ b/src/components/PropertyLookup.vue
@@ -51,10 +51,3 @@ export default Vue.extend( {
 	},
 } );
 </script>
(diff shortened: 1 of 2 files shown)


micgro42:
rebased on top of master

guergana (approved):
Lookin' good!

PR #133, opened 2021-01-07
Modify query and store to select or omit the labels

guergana:
added omitLabels to rootState and updated files that use the state

### DIFF
--- a/tests/unit/store/mutations.spec.ts
+++ b/tests/unit/store/mutations.spec.ts
@@ -138,6 +138,30 @@ describe( 'mutations', () => {
 
 	} );
 
+	it( 'setOmitLabels', () => {
+		const state: RootState = {
+			conditionRows: [
+				{
+					valueData: { value: 'potato', valueError: null },
+					propertyData: { id: 'P666', label: 'Day of the beast', datatype: 'string', propertyError: null },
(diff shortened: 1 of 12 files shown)


tests/unit/store/mutations.spec.ts
```diff
@@ -133,6 +138,30 @@ describe( 'mutations', () => {
...
+			],
+			limit: 0,
+			useLimit: false,
+			omitLabels: true,
+			errors: [],
+		};
+
+		const omitLabels = true;
+
+		mutations.setOmitLabels( state, omitLabels );
```
micgro42, line 159: This mutation is changing `omitLabels` from `true` to `true`. I think it would test more if one of them were false :)

tests/unit/store/mutations.spec.ts
```diff
@@ -133,6 +138,30 @@ describe( 'mutations', () => {
...
+
+		expect( state.omitLabels ).toStrictEqual( omitLabels );
```
micgro42: Since this is just a plain boolean, we can use `.toBe(` here instead of `.toStrictEqual(`

micgro42 (changes requested):
Solid work! One small comment about a test, then it is good to go!

micgro42 (approved):

PR #134, opened 2021-01-07
Show grey decorator around QueryCondition components

micgro42:
This is preparatory work for #131 . It wouldn't be strictly necessary yet, but it makes things much less confusing when comparing the QueryBuilder in #131 with the mocks.

This should be merged in master before #131

But: [T269470](https://phabricator.wikimedia.org/T270179)


guergana (approved):
looks good

PR #135, opened 2021-01-07
Add configuration for subclass property

micgro42:
This can differ on different Wikibase instances and so should be configurable.

Bug: [T271313](https://phabricator.wikimedia.org/T271313)


guergana (approved):
LGTM

PR #138, opened 2021-01-08
Make the margin for query wrapper follow the specs

Ladsgroup:
The specs says it should have xsmall margin on top and bottom:
https://www.figma.com/file/hZpaYX0xNNhK0L7kBvN6WK/Query-Builder?node-id=2604%3A29422

Currently, the add condition button is stuck to the wrapper border which
doesn't look nice


Ladsgroup:
Before this PR:
image

After this PR:
image

micgro42 (approved):

PR #141, opened 2021-01-08
Extract StringValueInput for better encapsulation

micgro42:
(This should be merged into master after #128 and is followed by #142)

This way the interfaces between StringValueInput and ItemValueLookup are nicely aligned which will make the next steps easier.


guergana (approved):
Looks good.

PR #143, opened 2021-01-08
Improve the box styles according to the specs

Ladsgroup:
- Improve responsiveness by removing the extra padding when the boxes
   are stacking on top of each other.
 - Add border based on the specs

### DIFF
--- a/src/components/QueryBuilder.vue
+++ b/src/components/QueryBuilder.vue
@@ -123,7 +123,9 @@ a {
 	margin-block: $dimension-layout-xsmall;
 	padding-block: $dimension-layout-xsmall;
 	padding-inline: $dimension-layout-xsmall;
-	background-color: $color-base-70; // maybe replace with an alias token?
+	background-color: $color-base-80; // maybe replace with an alias token?
+	border: $border-width-thin $border-style-base $border-color-base-default;
+	border-radius: $border-radius-base;
 }
 
 .querybuilder__heading {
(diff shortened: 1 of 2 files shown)


guergana:
Hi, Amir, where are the specs? I don't see the gray border in figma.

src/components/QueryBuilder.vue
```diff
@@ -123,7 +123,8 @@ a {
 	margin-block: $dimension-layout-xsmall;
 	padding-block: $dimension-layout-xsmall;
 	padding-inline: $dimension-layout-xsmall;
-	background-color: $color-base-70; // maybe replace with an alias token?
+	background-color: $color-base-80; // maybe replace with an alias token?
+	border: $border-width-thin $border-style-base $border-color-base-default;
```
micgro42, line 127: I think they still need the `$border-radius-base`

micgro42 (changes requested):
Thanks for those adjustments! Just the border radius is missing, I think :)

The specs have been added to here https://www.figma.com/file/hZpaYX0xNNhK0L7kBvN6WK/Query-Builder?node-id=2604%3A29420

Ladsgroup:
> Thanks for those adjustments! Just the border radius is missing, I think :)
> 
> The specs have been added to here https://www.figma.com/file/hZpaYX0xNNhK0L7kBvN6WK/Query-Builder?node-id=2604%3A29420

Oh thanks. They didn't exist in the CSS but they are added as a note (that's why I missed that)
image

micgro42 (approved):
Ready to go :)

PR #144, opened 2021-01-11
Layout adjustments for the tiniest viewport

micgro42:
This should be merged after #143 

Inspired by the styles in https://www.figma.com/file/hZpaYX0xNNhK0L7kBvN6WK/Query-Builder?node-id=2604%3A29422

The viewport breakpoint has been chosen so that the inputs don't align as next to each other again in the smallest layout.

Also, the outer padding was too large still even for larger viewports. It should be `-small`, not `-xlarge`.

Bug: [T269470](https://phabricator.wikimedia.org/T269470)

### DIFF
--- a/src/components/QueryBuilder.vue
+++ b/src/components/QueryBuilder.vue
@@ -114,9 +114,16 @@ a {
 	}
 }
 
+$tinyViewportWidth: 36em;
+
 .querybuilder {
-	padding-block: $dimension-spacing-xlarge;
-	padding-inline: $dimension-spacing-xlarge;
+	padding-block: $dimension-layout-small;
+	padding-inline: $dimension-layout-small;
+
+	@media (max-width: $tinyViewportWidth) {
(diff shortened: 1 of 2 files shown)


guergana (dismissed):
The outer spacing of the component in desktop are now gone.

micgro42:
> The outer spacing of the component in desktop are now gone.

Fixed now, I used a spacing token instead of a layout token.

Ladsgroup:
I might be doing it wrong but when I pulled the branch and ran it, it still got the error in 290px:
image

micgro42:
> I might be doing it wrong but when I pulled the branch and ran it, it still got the error in 290px:
> image

Yeah. that's ok. We don't support screens below 320px. We have to stop somewhere.

Ladsgroup:
Aha, I had the issue with 370px before and I thought it's the same. Missed the 100px difference.

Ladsgroup (approved):

PR #145, opened 2021-01-11
Modify query and store to show subclasses

guergana:
first commit: Modify store to accept subclasses
second commit: modify the query.
third commit: merge
fourth commit: this field should be unchecked by default and should be checked true automatically if and only if the user selects a wikibase-item property


tests/unit/sparql/buildQuery.spec.ts
```diff
@@ -162,10 +172,35 @@ describe( 'buildQuery', () => {
...
 		],
 		omitLabels: false,
 		} );
 		expect( actualQuery.replace( /\s+/g, ' ' ) ).toEqual( expectedQuery.replace( /\s+/g, ' ' ) );
 	} );
+
+	it( 'builds a query from a property and a string value with subclasses', () => {
+		const propertyId = 'P666';
+		const value = 'blah';
+		const subclassesId = 'P279';
```
guergana: How do we pass the environment variable here? I tried `const subclassesId = process.env.VUE_APP_SUBCLASS_PROPERTY;` and it didnt work

micgro42: You could try 

```
process.env = Object.assign(process.env, { VUE_APP_SUBCLASS_PROPERTY: 'value' });
```

( which I found there: https://github.com/vuejs/vue-test-utils/issues/193#issuecomment-438437626 )

guergana: Uf!! That was hard. The proposed solution wasn't working but this one worked: https://stackoverflow.com/a/58953365/1619792

package-lock.json
```diff
@@ -4194,19 +4194,19 @@
       "dependencies": {
         "ansi-regex": {
           "version": "2.1.1",
-          "resolved": "https://registry.npmjs.org/ansi-regex/-/ansi-regex-2.1.1.tgz",
+          "resolved": "https://registry.npm.taobao.org/ansi-regex/download/ansi-regex-2.1.1.tgz",
```
micgro42: We don't want those changes.

src/components/QueryCondition.vue
```diff
@@ -91,9 +91,10 @@ export default Vue.extend( {
 				if ( selectedProperty === null ) {
 					this.$store.dispatch( 'unsetProperty', this.conditionIndex );
 					return;
+				} else if ( selectedProperty.datatype === 'wikibase-item' ) {
```
micgro42: I don't think we need this `else` as the previous `if` always returns

src/components/QueryCondition.vue
```diff
@@ -91,9 +91,10 @@ export default Vue.extend( {
...
+					this.$store.dispatch( 'setSubclasses', { conditionIndex: this.conditionIndex, subclasses: true } );
```
micgro42, line 96: We do need to set this back to false when we select a different property, right?

package.json
```diff
@@ -77,5 +77,10 @@
   "repository": {
     "type": "git",
     "url": "git+https://github.com/wmde/query-builder.git"
+  },
+  "jest": {
+    "setupFiles": [
+      "./tests/unit/setEnvVars.js"
```
micgro42: Should that not be `.jest/setEnvVars.js`? I'm confused that the tests are still passing :thinking:

tests/unit/store/mutations.spec.ts
```diff
@@ -100,12 +103,40 @@ describe( 'mutations', () => {
...
+				conditionId: '0.123',
+				subclasses: true,
+			} ],
+			limit: 0,
+			useLimit: false,
+			omitLabels: true,
+			errors: [],
+		};
+
+		mutations.setSubclasses( state, { subclasses: expectedValue, conditionIndex: 0 } );
```
micgro42, line 128: This test sets the subclasses from `true` to `true`. Either the old or the new value should be `false` to actually test something.

micgro42 (changes requested):

package-lock.json
guergana: oh yes, missed that one

src/components/QueryCondition.vue
guergana: I wasn't sure if we needed it either.

package.json
guergana: Actually this was from another way I tried that didn't work. So this is actually not being used. Thanks for pointing it out.

src/components/QueryCondition.vue
guergana: I added an else here

tests/unit/store/mutations.spec.ts
guergana: ah, yes, good catch.

package-lock.json
```diff
@@ -4200,13 +4200,13 @@
         },
         "ansi-styles": {
           "version": "2.2.1",
-          "resolved": "https://registry.npmjs.org/ansi-styles/-/ansi-styles-2.2.1.tgz",
+          "resolved": "https://registry.npm.taobao.org/ansi-styles/download/ansi-styles-2.2.1.tgz",
```
micgro42: No, I mean, we don't want to have this file checked in at all for this PR (and also do not want to use registry.npm.taobao.org)

src/components/QueryCondition.vue
micgro42: :+1:

tests/unit/components/QueryCondition.spec.ts
```diff
@@ -102,6 +102,26 @@ describe( 'QueryCondition.vue', () => {
 		expect( store.dispatch ).toHaveBeenCalledWith( 'updateValue', { value: userInput, conditionIndex } );
 	} );
 
+	it( 'set subclasses to true when property type is wikibase-item', async () => {
```
micgro42, line 105: Can we have a test for setting it to false as well if the datatype is `string`?

micgro42 (changes requested):
Thanks! We are almost there :slightly_smiling_face:

package-lock.json
guergana: oh, weird. I thought I removed this.

micgro42 (approved):

PR #147, opened 2021-01-12
Debounce fetching resource in FetchSearchEntityRepository

Ladsgroup:
Bug: T267631

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,6 +25,7 @@
     "@wmde/wikit-tokens": "^2.0.0-alpha.2",
     "@wmde/wikit-vue-components": "^2.0.0-alpha.2",
     "core-js": "^3.6.5",
+    "lodash.debounce": "^4.0.8",
     "ress": "^3.0.0",
     "saucelabs": "^4.5.2",
     "sparqljs": "^3.0.3",
@@ -34,6 +35,7 @@
   },
   "devDependencies": {
     "@types/jest": "^24.0.19",
+    "@types/lodash": "^4.14.167",
     "@types/node": "^14.14.9",
     "@types/jest-axe": "^3.5.0",
     "@typescript-eslint/eslint-plugin": "^2.33.0",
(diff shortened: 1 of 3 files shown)


src/data-access/FetchSearchEntityRepository.ts
```diff
@@ -39,8 +40,12 @@ export default class FetchSearchEntityRepository implements SearchEntityReposito
 			url.searchParams.set( key, params[ key ] );
 		}
 		let response: Response;
+		const debouncedFetch = debounce( fetch, 300, {
+			leading: true,
+			trailing: false,
```
micgro42: No, I think we want the trailing edge. I.e. if I type four keys after each other, then I want to see the suggestions for the search result with four letters instead of one

micgro42 (changes requested):

package.json
```diff
@@ -22,9 +22,11 @@
   },
   "main": "index.js",
   "dependencies": {
+    "@types/lodash": "^4.14.167",
     "@wmde/wikit-tokens": "^2.0.0-alpha.2",
     "@wmde/wikit-vue-components": "^2.0.0-alpha.2",
     "core-js": "^3.6.5",
+    "lodash": "^4.17.20",
```
micgro42: also, we can use lodash.debounce here. That is what we're also doing in Wikit and that's then maybe a bit less of headache for the sec review :)

package.json
```diff
@@ -22,9 +22,11 @@
   },
   "main": "index.js",
   "dependencies": {
+    "@types/lodash": "^4.14.167",
```
micgro42: I think those types should go into the dev dependencies

src/data-access/FetchSearchEntityRepository.ts
Ladsgroup: Done

package.json
Ladsgroup: Done

package.json
Ladsgroup: Done

micgro42 (approved):

PR #148, opened 2021-01-13
feat: Add simple footer

micgro42:
The main purpose of this footer is to show the last build time, so we
can have an idea when the production environment is updated again.

Proper (translatable) copy will be implemented later.

Bug: [T268643](https://phabricator.wikimedia.org/T268643)

### DIFF
--- a/src/components/Footer.vue
+++ b/src/components/Footer.vue
@@ -1,0 +1,47 @@
+<template>
+	<footer class="querybuilder-footer">
+		<p>
+			Made by the Wikimedia Deutschland Wikidata team.
+			Licensed under <a href="https://github.com/wmde/query-builder/blob/master/LICENSE">BSD 3-Clause License</a>.
+			<a href="https://github.com/wmde/query-builder">View Source</a>.
+		</p>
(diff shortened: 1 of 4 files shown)


src/components/Footer.vue
```diff
@@ -0,0 +1,47 @@
+<template>
+	<footer class="querybuilder-footer">
+		<p>
+			Made by the Wikimedia Deutschland Wikidata team.
+			Licensed under <a href="https://github.com/wmde/query-builder/blob/master/LICENSE">BSD 3-Clause License</a>.
+			View on <a href="https://github.com/wmde/query-builder">GitHub</a>.
```
Ladsgroup: "View source" is a better word IMO.

micgro42: Will adjust

micgro42: Done

guergana:
I dont know what the current discussion about this is, but it looks good.

guergana (approved):
Looks good. I haven't participated in the discussions around the footer but looks great.

PR #149, opened 2021-01-13
Add placeholder to Limit component

micgro42:
This also moves the screen-reader label to the messages to make it translatable.

Bug: [T270712](https://phabricator.wikimedia.org/T270712)

### DIFF
--- a/public/i18n/en.json
+++ b/public/i18n/en.json
@@ -21,5 +21,7 @@
   "query-builder-value-type-relation-dropdown-regardless-of-value": "regardless of value",
   "query-builder-delete-condition": "Delete condition",
   "query-builder-limit-number-results-description": "Limit the number of results to",
+  "query-builder-limit-number-placeholder": "Enter a number",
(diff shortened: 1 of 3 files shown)


public/i18n/en.json
```diff
@@ -21,5 +21,7 @@
   "query-builder-value-type-relation-dropdown-regardless-of-value": "regardless of value",
   "query-builder-delete-condition": "Delete condition",
   "query-builder-limit-number-results-description": "Limit the number of results to",
+  "query-builder-limit-number-placeholder": "enter a number",
```
guergana: Is there a specific reason for this to be not to be capitalized? ( First letter in capitals? )

micgro42: That is how it was in https://phabricator.wikimedia.org/T270712 :shrug:, but maybe that was not intentional. I'll ask UX.

micgro42: Done. It is now uppercase.

guergana (approved):
LGTM

PR #150, opened 2021-01-14
Append property id to anyvalue to avoid clashes

Ladsgroup:
When several properties have been chosen with "regardless of value"

Bug: T271903

### DIFF
--- a/tests/unit/sparql/buildQuery.spec.ts
+++ b/tests/unit/sparql/buildQuery.spec.ts
@@ -55,14 +55,14 @@ describe( 'buildQuery', () => {
 			},
 		],
 		omitLabels: true,
-		} ) ).toEqual( `SELECT ?item WHERE { ?item (p:${propertyId}/ps:${propertyId}) _:anyValue. }` );
+		} ) ).toEqual( `SELECT ?item WHERE { ?item (p:${propertyId}/ps:${propertyId}) _:anyValue${propertyId}. }` );
 	} );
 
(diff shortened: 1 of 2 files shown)


tests/unit/sparql/buildQuery.spec.ts
```diff
@@ -55,14 +55,14 @@ describe( 'buildQuery', () => {
...
-		} ) ).toEqual( `SELECT ?item WHERE { ?item (p:${propertyId}/ps:${propertyId}) _:anyValue. }` );
+		} ) ).toEqual( `SELECT ?item WHERE { ?item (p:${propertyId}/ps:${propertyId}) _:anyValue${propertyId}. }` );
 	} );
 
 	it( 'builds a query from multiple conditions, one matching and one regardless', () => {
 		const expectedQuery =
 			`SELECT ?item WHERE {
 			?item (p:P666/ps:P666) "blah".
-			?item (p:P66/ps:P66) _:anyValue.
+			?item (p:P66/ps:P66) _:anyValueP66.
```
guergana, line 65: Can there be the case that a user searches for the same property in two different conditions? :thinking:

guergana (approved):

tests/unit/sparql/buildQuery.spec.ts
Ladsgroup: Currently no, we might have it once we have nesting but that requires a much bigger refactor.

PR #151, opened 2021-01-14
Create SubclassCheckbox component

guergana:
first commit: added translations
second commit: added component and test

### DIFF
--- a/src/components/SubclassCheckbox.vue
+++ b/src/components/SubclassCheckbox.vue
@@ -1,0 +1,44 @@
+<template>
+	<div class="querybuilder__include-subclasses">
+		<input
+			class="querybuilder__include-subclasses-checkbox"
+			type="checkbox"
+			id="include-subclasses"
+			:checked="isChecked"
+			@change="$emit( 'subclass-check', $event )"
+		/>
+		<label for="include-subclasses">
+			<span
+				v-i18n="{msg: 'query-builder-include-subclasses'}" />
+		</label>
+	</div>
+</template>
+
+<script lang="ts">
+import Vue from 'vue';
+export default Vue.extend( {
+	name: 'SubclassCheckbox',
+	props: {
+		isChecked: {
+			type: Boolean,
+			required: true,
+		},
+	},
+} );
+</script>
+
+<style lang="scss">
+.querybuilder__include-subclasses {
+	display: flex;
+	align-items: flex-end;
+	color: $font-color-base;
+	// TODO: change to real ones
+	font-family: $font-family-style-description;
+	margin-block-start: $dimension-layout-medium;
+
+	&-checkbox {
+		// TODO: change to real ones
+		margin-inline-end: $dimension-layout-xxsmall;
+	}
+}
+</style>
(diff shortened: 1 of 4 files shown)


src/components/SubclassCheckbox.vue
```diff
@@ -0,0 +1,47 @@
...
+import Vue from 'vue';
+export default Vue.extend( {
+	name: 'SubclassCheckbox',
+	computed: {
+		checked: {
+			get(): boolean {
+				return this.$store.getters.subclasses;
+			},
+			set( value: boolean ): void {
+				this.$store.dispatch( 'setSubclasses', value );
```
micgro42: No, this is missing the info about the conditionIndex. In general, I think it would be the better idea if this component doesn't talk to the store at all and just has a value prop and emits an input event when the checkbox is toggled. Let's let `QueryCondition` deal with the talking to the store.

micgro42 (changes requested):

src/components/SubclassCheckbox.vue
guergana: oh, true.. :sweat: totally forgot about the index.

src/components/SubclassCheckbox.vue
```diff
@@ -0,0 +1,37 @@
+<template>
+	<div class="querybuilder__include-subclasses">
+		<input
```
micgro42, line 3: Now we need a way to to actually set the checked state from the outside. I.e. some boolean prop (e.g. `value` or `isChecked`) that then is assigned to the `checked` attributed of the input :slightly_smiling_face: .

micgro42 (changes requested):
Almost there, I think. Just one prop missing

src/components/SubclassCheckbox.vue
guergana: you mean for when it's coming from the store at the start?

micgro42: Yes, exactly

guergana: gotcha.

src/components/SubclassCheckbox.vue
```diff
@@ -0,0 +1,43 @@
...
+</template>
+
+<script lang="ts">
+import Vue from 'vue';
+export default Vue.extend( {
+	name: 'SubclassCheckbox',
+	props: {
+		isChecked: {
+			type: Boolean,
+			required: true,
```
guergana, line 24: it will always be in the store. so i guess it makes sense to make it required.

src/components/QueryCondition.vue
```diff
@@ -6,8 +6,7 @@
 			@click="removeCondition"
 		/>
 		<div class="query-condition__toggle-button-group">
-			<Button disabled="true" style="border-right: 1px darkgrey solid;">with</Button>
-			<Button disabled="true">without</Button>
+			<NegationToggle value="without" />
```
micgro42: I think this was added accidentally :)

micgro42 (changes requested):

src/components/QueryCondition.vue
guergana: ugh! yes.

micgro42 (approved):

PR #152, opened 2021-01-14
Add NegationToggle component

micgro42:
The tests will be added when we start using the ButtonToggle component from the Design System.

Bug: [T272011](https://phabricator.wikimedia.org/T271788)

### DIFF
--- a/src/components/NegationToggle.vue
+++ b/src/components/NegationToggle.vue
@@ -1,0 +1,36 @@
+<template>
+	<span class="negationToggle">
+		<Button
+			:type="value==='with' ? 'progressive' : 'neutral'"
+			:variant="value==='with' ? 'primary' : 'normal'"
+			@click.native="$emit('input', 'with')"
+		>{{ $i18n( 'query-builder-negation-option-label-with' ) }}</Button>
+		<Button
+			:type="value==='without' ? 'progressive' : 'neutral'"
+			:variant="value==='without' ? 'primary' : 'normal'"
+			@click.native="$emit('input', 'without')"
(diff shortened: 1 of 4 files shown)


src/components/NegationToggle.vue
```diff
@@ -0,0 +1,36 @@
+<template>
+	<span class="negationToggle">
+		<Button
+			:type="value==='with' ? 'progressive' : 'neutral'"
```
guergana, line 4: it doesn't seem to change to progressive (as in blue background). when it is selected it only shows a blue border around the selected button. is this the desired effect?

src/components/NegationToggle.vue
```diff
@@ -0,0 +1,36 @@
...
+			:variant="value==='with' ? 'primary' : 'normal'"
+			@click.native="$emit('input', 'with')"
+		>{{ $i18n( 'query-builder-negation-option-label-with' ) }}</Button>
+		<Button
+			:type="value==='without' ? 'progressive' : 'neutral'"
+			:variant="value==='without' ? 'primary' : 'normal'"
```
guergana, line 10: same as previous comment. doesn't change the class. the button is always normal

guergana (dismissed):
Doesnt change the class when one of the buttons is clicked.

guergana (commented):
Looks good. Tho we might be relegating a lot into the QueryCondition Component.

guergana:
Has to be rebased and I think it's good to go. :)

guergana (approved):

micgro42:
Rebased on top of master :+1:

PR #155, opened 2021-01-15
Use Subclass Checkbox

guergana:
Not sure if more tests are needed. Seems they are all included in the previous commit.

### DIFF
--- a/src/components/QueryCondition.vue
+++ b/src/components/QueryCondition.vue
@@ -31,6 +31,10 @@
 					:error="valueError"
 					:datatype="datatype"
 				/>
+				<SubclassCheckbox
+					v-if="subclassCheckboxVisible"
+					@subclass-check="setSubclasses"
+					:isChecked="subclasses(conditionIndex)" />
 			</div>
 			<div>
 				<Dropdown
@@ -52,6 +56,7 @@ import ValueInput from '@/components/ValueInput.vue';
 import DeleteConditionButton from '@/components/DeleteConditionButton.vue';
 import PropertyLookup from '@/components/PropertyLookup.vue';
 import ValueTypeDropDown from '@/components/ValueTypeDropDown.vue';
+import SubclassCheckbox from '@/components/SubclassCheckbox.vue';
 import SearchResult from '@/data-access/SearchResult';
 import PropertyValueRelation from '@/data-model/PropertyValueRelation';
 import Error from '@/data-model/Error';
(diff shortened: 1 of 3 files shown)


src/components/SubclassCheckbox.vue
```diff
@@ -5,7 +5,7 @@
 			type="checkbox"
 			id="include-subclasses"
 			:checked="isChecked"
-			@change="$emit( 'subclass-check', $event )"
+			@change="$emit( 'subclass-check', $event.target.checked )"
```
guergana, line 8: We don't need the whole event. We just just emit what we need.

src/components/QueryCondition.vue
```diff
@@ -32,6 +32,7 @@
 					:error="valueError"
 					:datatype="datatype"
 				/>
+				<SubclassCheckbox @subclass-check="setSubclasses" :isChecked="subclasses(conditionIndex)" />
```
micgro42: This should only be shown for properties with datatype `wikibase-item`

src/components/SubclassCheckbox.vue
```diff
@@ -5,7 +5,7 @@
 			type="checkbox"
 			id="include-subclasses"
```
micgro42: Now all these checkboxes have the same id, which means if I have two conditions and I click the label of the checkbox in the second condition, then the checkbox in the first condition is toggled.

micgro42 (changes requested):

src/components/SubclassCheckbox.vue
micgro42: Something like `:id="id"`
and 
```js
data() {
    return {
        id: `include-subclasses-${Math.random()}`,
    };
},
```
should be good enough

src/components/QueryCondition.vue
guergana: Oh, I understood it was only checked when datatype is `wikibase-item`  but always shown? Will update.

guergana: addressed!

src/components/SubclassCheckbox.vue
guergana: addressed!

src/components/SubclassCheckbox.vue
```diff
@@ -3,9 +3,9 @@
 		<input
 			class="querybuilder__include-subclasses-checkbox"
 			type="checkbox"
-			id="include-subclasses"
+			:id="id"
 			:checked="isChecked"
-			@change="$emit( 'subclass-check', $event )"
+			@change="$emit( 'subclass-check', $event.target.checked )"
 		/>
 		<label for="include-subclasses">
```
micgro42: You have to use the dynamic id here as well :)

micgro42 (changes requested):
Almost there!

micgro42 (approved):
\o/

PR #156, opened 2021-01-15
Implement negation

Ladsgroup:
- Changing the state to accept negation in its mutation, getter, and
   actions.
 - Using the negation toggle component and make it send the change to
   state (and read from it, duh)
 - Rework state tests to have a base RootState and change that so for
   every change of state we wouldn't need to do the change in every case
 - Add tests for state changes and query condition using the negation

Bug: T272014
Bug: T272015


src/components/QueryCondition.vue
```diff
@@ -5,10 +5,10 @@
 			:disabled="!canDelete"
 			@click="removeCondition"
 		/>
-		<div class="query-condition__toggle-button-group">
-			<Button disabled="true" style="border-right: 1px darkgrey solid;">with</Button>
-			<Button disabled="true">without</Button>
-		</div>
+		<NegationToggle
+			class="query-condition__toggle-button-group"
```
micgro42, line 9: We not using this class anywhere, right? If so, then we can just drop it

src/components/QueryCondition.vue
```diff
@@ -122,6 +123,24 @@ export default Vue.extend( {
 				this.$store.dispatch( 'updateValue', { value, conditionIndex: this.conditionIndex } );
 			},
 		},
+		negate: {
```
micgro42: You can do that with one computed property. Having those two seems a bit too much

tests/unit/store/getters.spec.ts
```diff
@@ -3,97 +3,53 @@ import getters from '@/store/getters';
...
-					propertyValueRelationData: { value: PropertyValueRelation.Matching },
-					conditionId: '0.123',
-					subclasses: false,
-				} ],
-				omitLabels: true,
-				errors: [],
-				limit: 0,
-				useLimit: false,
-			};
+			const state: RootState = JSON.parse( JSON.stringify( simpleRootState ) );
```
micgro42, line 31: That could be done nicer with a method `getFreshRootState(): RootState`, but that can be refactored later. This is already an improvement :pray:

micgro42 (changes requested):
Good start so far. In addition to the two points below, we can also make the `negate` field non-optional in the QueryRepresentation

src/components/QueryCondition.vue
Ladsgroup: It's used (at least for now, until we use the wikit component, look at `&__toggle-button-group` line 190.

micgro42: Narf! I totally missed that! Cool, then that class is fine right now. Its naming is not ideal, but that can be changed in another PR, not this one.

src/components/QueryCondition.vue
Ladsgroup: Yeah, it grew organically when I was writing the code. Fixed now.

Ladsgroup:
> d start so far. In addition to the two points below, we can also make the `negate` field non-optional in the QueryRepresentation

I just tried it but buildQuery.js tests exploded, needs a refactor that I'll do separately with the `getFreshRootState()` refactor.

src/components/QueryCondition.vue
micgro42: :+1:

micgro42 (approved):

PR #157, opened 2021-01-18
fix: align the icon so that it fits properly into the Button

micgro42:
However, for this patch to take affect, the vertical-align rule on the icon in the Wikit Design System has to be removed.

See https://stackoverflow.com/questions/65741031/what-influences-the-actually-rendered-inner-size-of-an-inline-element

Bug: [T268982](https://phabricator.wikimedia.org/T268982)


Ladsgroup:
It's good to go, but let's wait for the wikit to be merged first and then it has my virtual approval.

micgro42:
Rebased it on top of master

Ladsgroup (approved):

PR #158, opened 2021-01-18
Ensure margin between AddCondition and Limit components

micgro42:
Floating right is ignoring margins, so we use flex instead so the AddCondition button takes up the space it needs.

Bug: [T270712](https://phabricator.wikimedia.org/T270712#6747720)


bereket-WMDE (approved):
tested on FF and chrome.
It looks good!

PR #159, opened 2021-01-18
Fix bug in negation

Ladsgroup:
When checking if the "where" condition has only negation or not, if
there is the service for labeling (which is on by default), it thinks
the query has a non-negate value and avoids adding "all items" bit
rendering the query useless.


micgro42 (changes requested):
Good catch, thank you! Can we get a test for this? Or maybe adjust an existing test to account for this?

Ladsgroup:
> Good catch, thank you! Can we get a test for this? Or maybe adjust an existing test to account for this?

Good idea. Thanks. Done.

micgro42 (approved):

PR #160, opened 2021-01-18
Rework tests to make negate required in QueryRepresentation

Ladsgroup:
Refactoring the tests so it would produce a simple condition from
property and value and then overriding the defaults if needed.

Also, cleaning up other test to use a function as well.


tests/unit/sparql/buildQuery.spec.ts
```diff
@@ -1,63 +1,62 @@
...
+		const simpleCondition: Condition = {
+			propertyId,
+			value,
+			datatype: 'string',
+			propertyValueRelation,
+			subclasses: false,
+			negate: false,
+		};
+
+		return JSON.parse( JSON.stringify( simpleCondition ) );
```
micgro42: This JSON.parse and JSON.stringify should not be necessary. I think this method always returns a fresh now object, not a reference to the same.

tests/unit/store/getters.spec.ts
```diff
@@ -3,44 +3,46 @@ import getters from '@/store/getters';
...
+				conditionId: '0.123',
+				subclasses: false,
+				negate: false,
+			} ],
+			limit: 0,
+			useLimit: false,
+			omitLabels: true,
+			errors: [],
+		};
+		return JSON.parse( JSON.stringify( simpleRootState ) );
```
micgro42: Also that JSON.parse and JSON.stringify shouldn't be needed. Just return the object :)

micgro42 (changes requested):
We can make it a bit simpler :)

Ladsgroup:
> We can make it a bit simpler :)

Done. Thanks!

micgro42 (approved):

PR #163, opened 2021-01-19
Add serialization and deserialization for sharing Url

guergana:
first commit: serializer
second commit: deserializer

### DIFF
--- a/src/serialization/QuerySerializer.ts
+++ b/src/serialization/QuerySerializer.ts
@@ -1,0 +1,26 @@
+import RootState, { ConditionRow, ItemValue } from '@/store/RootState';
+import SerializedObject from '@/data-model/SerializedObject';
+
+export default class QuerySerializer {
+	public serialize( state: RootState ): string {
+		const conditions: SerializedObject[] = [];
+		state.conditionRows.forEach( ( condition: ConditionRow ) => {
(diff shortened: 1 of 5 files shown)


src/serialization/QuerySerializer.ts
```diff
@@ -0,0 +1,10 @@
+import QueryRepresentation from '@/sparql/QueryRepresentation';
+
+export default class QuerySerializer {
+	public serialize( query: QueryRepresentation ): object {
```
micgro42: That serializer should probably return a string, not an object.

micgro42: I'm sorry, but you can't start with a `QueryRepresentation`. The `query` getter (rightly) assumes an already valid Query Builder, but we want to be able to generate this link even with just a partially filled query builder.
So the input should be the root state itself. That being said, it is probably enough to serialize the same information that will end up in the `QueryRepresentation`, i.e. no need to serialize errors or labels or the `conditionId`.

guergana: Ok! Makes sense too.

guergana: Adressed

Ladsgroup (approved):

PR #165, opened 2021-01-20
Create results header and frame

Ladsgroup:
Bug: T271794

### DIFF
--- a/src/components/QueryResult.vue
+++ b/src/components/QueryResult.vue
@@ -1,22 +1,28 @@
 <template>
 	<div class="querybuilder__result">
+		<div class="querybuilder__result__header">
+			<h2>{{ $i18n('query-builder-result-header')}}</h2>
+		</div>
 		<div v-if="errors.length !== 0" class="querybuilder__result__errors">
 			<Message
 				v-for="(error, index) in errors" :type="error.type" :key="index">
 				<span>{{$i18n(error.message)}}</span>
 			</Message>
 		</div>
-		<div v-else-if="encodedQuery.length === 0" class="querybuilder__result__description">
-			{{ $i18n('query-builder-result-placeholder')}}
+		<div v-else-if="encodedQuery.length === 0">
+			<div class="querybuilder__result__description">
+				{{ $i18n('query-builder-result-placeholder')}}
+			</div>
 		</div>
-		<iframe
-			v-else
-			:src="queryServiceEmbedUrl + '#' + encodedQuery"
-			:key="iframeRenderKey"
-			class="querybuilder__result__iframe"
-			referrerpolicy="origin"
-			sandbox="allow-scripts allow-same-origin allow-popups">
-		</iframe>
+		<div v-else class="querybuilder__result__wrapper">
+			<iframe
+				:src="queryServiceEmbedUrl + '#' + encodedQuery"
+				:key="iframeRenderKey"
+				class="querybuilder__result__iframe"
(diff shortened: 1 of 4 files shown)


Ladsgroup:
image

image


Stuff to double check with UX before going in:
 - Border radius
 - Height
 - Font of header
 - Margin of the frame

src/components/QueryResult.vue
```diff
@@ -9,14 +9,18 @@
...
-			v-else
-			:src="queryServiceEmbedUrl + '#' + encodedQuery"
-			:key="iframeRenderKey"
-			class="querybuilder__result__iframe"
-			referrerpolicy="origin"
-			sandbox="allow-scripts allow-same-origin allow-popups">
-		</iframe>
+		<div v-else class="querybuilder__result__wrapper">
+			<div class="querybuilder__result__header">
+				<h5>{{ $i18n('query-builder-result-header')}}</h5>
```
micgro42: This should be an `<h2>`, though it might need some more styling

Ladsgroup: The figma spec says heading 5, I'm slightly confused about this now.

micgro42: I think that might be referring to a set of styles dubbed "heading 5", but lets ask to be on the safe side

Ladsgroup:
Styles are a bit messy but I think the result basically is what UX wanted.

src/components/QueryResult.vue
```diff
@@ -70,12 +79,38 @@ export default Vue.extend( {
...
+		width: $dimension-width-full;
+		border: $border-width-thin $border-style-base $border-color-base-subtle;
+		border-radius: 0 0 $border-radius-base $border-radius-base;
 		text-align: center;
 	}
 
+	.querybuilder__result__header {
+		display: flex;
+		align-items: center;
+		background: $background-color-neutral-default;
```
sai-san:
```suggestion
		background: $color-base-90;
```

sai-san: I believe that we should use a global token here, since that "neutral" background is semantically related to the "flavor" of some components (e.g. buttons). The decision can be revised in the event that this header is turned into a component.

src/components/QueryResult.vue
```diff
@@ -70,12 +79,38 @@ export default Vue.extend( {
...
+			font-weight: $font-weight-style-h5;
+			font-size: $font-size-style-h5;
+			color: $font-color-base;
+		}
+	}
+
 	.querybuilder__result__iframe {
 		width: $dimension-width-full;
-		border: none;
+		background: $background-color-neutral-default;
```
sai-san:
```suggestion
		background: $color-base-90;
```

sai-san: 👆🏻 same as above

src/components/QueryResult.vue
Ladsgroup: That was mentioned in the spec but if that's what you think should be, I'm all fine with it.

src/components/QueryResult.vue
Ladsgroup: Oh this one doesn't even need it. We can't set background of the iframe (cleaning up the result css is another thing altogether which I actually love to do)

sai-san: yep, I wondered about it! Thanks, Amir! This is looking excellent 💯

micgro42 (approved):

PR #167, opened 2021-01-25
Rework labelling to improve its performance

Ladsgroup:
Basically wrapping the actual query with the labelling service, see the
discussion in T212933 and T166139

Bug: T272490


guergana (approved):
Looks good too.

PR #170, opened 2021-01-27
Modified the store to handle conditionRelations

guergana:


Ladsgroup:
Your package-lock.json is now exploded. You really like the Chinese npm.

guergana:
> Your package-lock.json is now exploded. You really like the Chinese npm.

ugh! gross!! will remove asap.

src/store/actions.ts
```diff
@@ -97,6 +98,11 @@ export default ( searchEntityRepository: SearchEntityRepository, metricsCollecto
 		payload: { subclasses: boolean; conditionIndex: number } ): void {
 		context.commit( 'setSubclasses', payload );
 	},
+	setConditionRelation(
+		context: ActionContext<RootState, RootState>,
+		payload: { value: ConditionRelation; conditionIndex: number } ): void {
```
Ladsgroup: I think this should also accept null.

src/store/mutations.ts
```diff
@@ -40,14 +41,24 @@ export default {
...
 	},
+	setConditionRelation( state: RootState, payload: { value: ConditionRelation; conditionIndex: number } ): void {
+		state.conditionRows[ payload.conditionIndex ].conditionRelation = payload.value;
+	},
 	setErrors( state: RootState, errors: Error[] ): void {
 		state.errors = errors;
 	},
 	addCondition( state: RootState ): void {
-		state.conditionRows.push( getFreshConditionRow() );
+		if ( state.conditionRows.length === 0 ) {
```
Ladsgroup: This can be easily summarized:
`state.conditionRows.push( getFreshConditionRow( state.conditionRows.length === 0 ) );`

src/data-model/SerializedObject.ts
```diff
@@ -1,11 +1,13 @@
...
 import { Value } from '@/store/RootState';
+import ConditionRelation from './ConditionRelation';
 
 export default interface SerializedCondition {
 	propertyId: string;
 	propertyDataType: string | null;
 	propertyValueRelation: PropertyValueRelation;
 	value: string | Value;
 	subclasses: boolean;
+	conditionRelation: ConditionRelation | null;
```
Ladsgroup, line 11: The logic in the code implies this condition relation is between this condition and the condition above (funnily enough, when implementing the component, I went with the exact opposite). This should be documented.

Ladsgroup (changes requested):

src/store/mutations.ts
guergana: wait what???

guergana: I think that is not very easy to read, IMHO

src/data-model/SerializedObject.ts
guergana: You mean by adding a comment? Or documented how?

src/store/actions.ts
guergana: agree! :)

src/store/mutations.ts
guergana: On second thoughts.. yes, it looks better.

src/data-model/SerializedObject.ts
Ladsgroup: Yes. Technically the relation doesn't belong to the query condition. It determines the relation between two. A comment in the QueryCondition would be good enough IMO

guergana: I added some comments here

Ladsgroup: Thanks. It's just it's below it and it should be usually above. Also maybe it should be in the RootState? here is mostly about sharing the link.

guergana: done. :roll_eyes:

Ladsgroup (approved):
Thanks!

PR #171, opened 2021-01-27
Set max width for entire UI

bereket-WMDE:
[Bug:T272055](https://phabricator.wikimedia.org/T272055)

### DIFF
--- a/src/components/QueryBuilder.vue
+++ b/src/components/QueryBuilder.vue
@@ -117,6 +117,12 @@ a {
 
 $tinyViewportWidth: 36em;
 
+/*
+ * This is a rough approximation. by using a screen size emulator,
+ * we can see the width where all items are aligned.
+ */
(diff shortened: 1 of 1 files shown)


src/components/QueryBuilder.vue
```diff
@@ -117,6 +117,11 @@ a {
 
 $tinyViewportWidth: 36em;
 
+/* This is a rough approximation. by using a screen size emulator, we can see the
+*width where all items are aligned.
```
Ladsgroup: Can this be something like this?
```
/*
 * Foo.
 * Bar.
 */
```

bereket-WMDE: updated

Ladsgroup (approved):

src/components/QueryBuilder.vue
Ladsgroup: Thanks!

PR #172, opened 2021-01-27
Swap the condition if it's negate only

Ladsgroup:
Otherwise, it can't find anything.

### DIFF
--- a/src/sparql/QueryObjectBuilder.ts
+++ b/src/sparql/QueryObjectBuilder.ts
@@ -42,7 +42,7 @@ export default class QueryObjectBuilder {
 			}
 
 			if ( isOnlyNegateQuery ) {
-				this.queryObject.where.push(
+				this.queryObject.where.unshift(
 					{
 						type: 'bgp',
 						triples: [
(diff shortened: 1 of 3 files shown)


micgro42:
We can ignore Sonar on this PR - it is from an old configuration.

But I'm confused why this PR is needed at all. With the QB on the tests-system I can find results just fine with only a single without clause?

Ladsgroup:
It happens in a rather specific case.
image

(Reported by Guergana)

The resulting query is https://w.wiki/w5z
Swap it and it'll works.

micgro42:
Mh, the problem seems to be the labels. Without the labels it works just fine for me:

image
https://w.wiki/wZb

With labels it reliably times out, even if I set the limit to 2. Does it actually work for you with labels?

Ladsgroup:
Yeah but the thing is that without the patch it gives "no result" which is semantically incorrect, while timeout is an operational issue (e.g. I hope it'll get better with the search team addressing some issues.)

micgro42 (approved):
Oh well. I'll create a ticket to figure out why the labels always time out.

PR #173, opened 2021-01-27
Introduce ConditionRelationToggle and use it in QueryCondition

Ladsgroup:
Currently, since the sparql part not merged yet, this won't work but it fully works with the state.

Bug: T272981

### DIFF
--- a/src/components/QueryBuilder.vue
+++ b/src/components/QueryBuilder.vue
@@ -12,17 +12,33 @@
 					v-i18n="{msg: 'query-builder-find-all-items'}" />
 				<div v-if="!conditionRows.length"
 					class="querybuilder__condition-placeholder"
-					v-i18n="{msg: 'query-builder-condition-placeholder'}"
-				/>
+					v-i18n="{msg: 'query-builder-condition-placeholder'}"/>
 				<div
-					class="querybuilder__condition-wrapper"
 					v-for="(condition, index) in conditionRows"
 					:key="condition.conditionId"
+					:class="[
+						'querybuilder__condition-group',
+						(isAboveOr(index)) ? 'querybuilder__condition-group-above' : '',
+					]"
 				>
-					<QueryCondition
+					<div
+						:class="[
+							'querybuilder__condition-wrapper',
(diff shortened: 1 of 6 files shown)


tests/unit/components/QueryBuilder.spec.ts
```diff
@@ -58,4 +60,36 @@ describe( 'QueryBuilder.vue', () => {
 		const wrapper = shallowMount( QueryBuilder, { store, localVue } );
 		expect( wrapper.find( '.querybuilder__condition-placeholder' ).exists() ).toBeFalsy();
 	} );
+
+	it( 'shows the "or" in toggle when there is two conditions', () => {
```
guergana: a wording thingy: 
shows the "or" in toggle when there ARE two conditions

guergana (changes requested):

tests/unit/components/QueryBuilder.spec.ts
Ladsgroup: oh right. Done.

src/components/QueryBuilder.vue
```diff
@@ -168,6 +197,30 @@ $largeViewportWidth: 90em; //~1438px
...
+	border-inline: $border-width-thin $border-style-base $border-color-base-default;
+	background-color: $color-base-80;
+}
+
+.querybuilder__condition-relation-toggle-or {
+	padding-inline-start: $dimension-layout-xsmall;
+}
+
+.querybuilder__condition-relation-toggle {
+	padding-block: $dimension-layout-xsmall;
```
guergana, line 221: I am testing it and I see the and has the spacing and the or is on the left. I think the logic is inverted. 

image

guergana (changes requested):

src/components/QueryBuilder.vue
Ladsgroup: Ugh. It was correct but once I was done, I did some refactoring that broke it. It should be fixed now.

guergana (approved):
Looks really cool

PR #174, opened 2021-01-28
Add the link that opens the query in WDQS

Ladsgroup:
Also, reworking the styles of <a> a bit

Bug: T272682


micgro42 (approved):
Works and the code looks good :+1:

PR #176, opened 2021-01-29
ci: track code coverage with sonar

micgro42:
Sonar is also used/evaluated by the foundation, e.g. for Wikibase ([Gerrit](https://gerrit.wikimedia.org/r/c/mediawiki/extensions/Wikibase/+/656958/3#message-2e9733f60b91084b9423aced9cfafb9710e0dce4), [Sonarcloud.io](https://sonarcloud.io/dashboard?id=mediawiki-extensions-Wikibase)).

### DIFF
--- a/.github/workflows/test-and-lint.yml
+++ b/.github/workflows/test-and-lint.yml
@@ -16,3 +16,26 @@ jobs:
       run: npm ci
     - name: Test
       run: npm run test
+
+  unit-test-with-coverage-and-sonar:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v2
+    - name: Set up node
+      uses: actions/setup-node@v1
+      with:
+        node-version: '15.x'
+    - name: Install
+      run: npm ci
+    - name: Test
+      run: npm run test:unit -- --coverage
+    - name: Setup Sonar Scanner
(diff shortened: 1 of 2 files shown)


.github/workflows/test-and-lint.yml
```diff
@@ -15,4 +15,27 @@ jobs:
...
+
+  unit-test-with-coverage-and-sonar:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v2
+    - name: Set up node
+      uses: actions/setup-node@v1
+      with:
+        node-version: '15.x'
```
bereket-WMDE, line 28: should the node-version be a variable ? 
would having different node-versions (line 14) be an issue ?

micgro42: Node 10 is severely outdated, but it is what is still used by our docker container, I think. I'm all in favor of upgrading that everywhere to a more recent version.

However, I would do that in a different PR. I chose 15 here mainly because that is what I used in my own project so far and I don't see any downside in using it. If at all, it gives us more confidence for the migration mentioned above.

bereket-WMDE: I see. definitely the upgrade should be a different PR.

bereket-WMDE (approved):

PR #178, opened 2021-01-29
refactor: extract setup and helper methods in tests

micgro42:
This is inspired by the pattern seen in bridge: https://github.com/wikimedia/Wikibase/blob/master/client/data-bridge/tests/util/store.ts


Ladsgroup (approved):
Usually I don't like utils/ as they slowly become a kitchen sink of incohesive things but given that this is tests and a rather small product. I don't see the point not to

PR #179, opened 2021-01-29
refactor: listen for event on ToggleButtonGroup not ToggleButtons

micgro42:
Listening on the `ToggleButtonGroup` is the way it is intended to be used as can be seen in the Wikit Storybook. Listening on the buttons creates a lot of redundancy if there are many buttons.

Also this PR adds the tests for this component.


Ladsgroup (approved):
Oh that's clever. I was struggling to make it work and couldn't find a good way (maybe I missed something in the documentation?) Thanks!

PR #180, opened 2021-01-29
ci: add info about previous commits to Sonar

micgro42:
By default actions/checkout@v2 only fetches a single commit. It would be nice to have the entire commit history available in Sonar.

Also, it makes that small warning go away in the Sonar interface, which is also nice.


Ladsgroup (approved):
I confirm: https://github.com/actions/checkout

This might get a little slow in the far future. Will think about it once it arises.

PR #183, opened 2021-02-01
Splitting TripleBuilder class from QueryObjectBuilder

Ladsgroup:
It cuts the class to roughly half and also makes it much more
understandable.


micgro42 (approved):
Loving it!

PR #184, opened 2021-02-01
Always show the link to WDQS

Ladsgroup:
As pointed out by UX in T272682#6792149

Bug: T272682


micgro42 (approved):
Unexpected, but I think it does what UX wants it to do.

PR #185, opened 2021-02-01
Wire referenceRelation to store

Ladsgroup:
Bug: T272699

### DIFF
--- a/src/components/QueryCondition.vue
+++ b/src/components/QueryCondition.vue
@@ -129,8 +129,13 @@ export default Vue.extend( {
 		},
 		selectedReferenceRelation: {
 			get(): ReferenceRelation {
-				// TODO: Implement
-				return ReferenceRelation.Regardless;
+				return this.$store.getters.referenceRelation( this.conditionIndex );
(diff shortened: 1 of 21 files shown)


src/components/QueryCondition.vue
```diff
@@ -41,7 +41,6 @@
 				<ReferenceRelationDropDown
 					class="query-condition__references"
 					v-model="selectedReferenceRelation"
-					:disabled="true"
```
micgro42: This doesn't actually work yet, right? In the sense of producing the correct SPARQL? Then I would say let's keep it disabled for now until the user can actually use it.

micgro42 (changes requested):
I think that enabling that dropdown in this PR might be a bit premature. But maybe I'm missing something?

src/components/QueryCondition.vue
Ladsgroup: Yeah, I enabled it to test it but forgot to disable it back again.

micgro42 (approved):
\o/

PR #186, opened 2021-02-02
Modify query to include conditionRelation unions

guergana:


micgro42 (approved):
:+1: Next we can do some refactoring

PR #187, opened 2021-02-03
Implement filtering by references

guergana:
first commit: only with references.
second commit: `without` reference case.
third commit: added missing triple.
fourth commit: enable ReferenceRelationDropdown

### DIFF
--- a/src/sparql/QueryObjectBuilder.ts
+++ b/src/sparql/QueryObjectBuilder.ts
@@ -1,6 +1,6 @@
 import rdfNamespaces from '@/sparql/rdfNamespaces';
 import QueryRepresentation, { Condition } from '@/sparql/QueryRepresentation';
-import { FilterPattern, Pattern, SelectQuery, Variable } from 'sparqljs';
+import { FilterPattern, Pattern, SelectQuery, Triple } from 'sparqljs';
 import PropertyValueRelation from '@/data-model/PropertyValueRelation';
 import TripleBuilder from '@/sparql/TripleBuilder';
 import ConditionRelation from '@/data-model/ConditionRelation';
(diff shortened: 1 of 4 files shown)


src/sparql/TripleBuilder.ts
```diff
@@ -101,6 +103,23 @@ export default class TripleBuilder {
 		return items;
 	}
 
+	public buildTripleForFilter( condition: Condition ): Triple {
```
Ladsgroup: Looking at the triple, it doesn't look like it's a triple for Filter. It looks like a reference triple ("wasDerivedFrom"). What about naming it "buildReferenceTriple"?

src/sparql/TripleBuilder.ts
```diff
@@ -2,13 +2,15 @@ import { Condition } from '@/sparql/QueryRepresentation';
...
+import ReferenceRelation from '@/data-model/ReferenceRelation';
 
 export default class TripleBuilder {
 	public buildTripleFromQueryCondition( condition: Condition ): Triple {
 		return {
 			subject: {
 				termType: 'Variable',
-				value: 'item',
+				value: condition.referenceRelation === ReferenceRelation.Regardless ?
+					'item' : 'statement' + condition.propertyId,
```
Ladsgroup: Technically it's correct but we can have a property being repeated more than once (occupation matching artist and occupation matching engineer) so maybe adding condition index would be better?

guergana: Yes, agree. I mentioned it at the top of the conversation here. That's one of the things that are still missing. Thanks for the suggestion.

src/sparql/TripleBuilder.ts
guergana: I just named it the first thing that came to my mind, yep `buildReferenceTriple` sounds better, will update.

src/sparql/QueryObjectBuilder.ts
```diff
@@ -123,11 +123,19 @@ export default class QueryObjectBuilder {
 			triples: [ this.tripleBuilder.buildTripleFromQueryCondition( condition, conditionIndex ) ],
 		};
 
-		const existsReferenceFilter: FilterPattern = {
+		let referenceExistsOrNot = '';
```
guergana, line 147: Not 100% sure about this. If I don't initialize it the `expression` in `referenceFilter` complains. Open to suggestions here.

Ladsgroup: I think it's good TBH

Ladsgroup (approved):

src/sparql/QueryObjectBuilder.ts
guergana: Haha. Can't believe this is approved. (Next comes refactoring)

PR #188, opened 2021-02-08
Split PatternBuilder object out of QueryObjectBuilder

Ladsgroup:
To make the object smaller and more understandable. It's just moving
around.

Bug: T272980


micgro42 (approved):
I like it.
I think my original goal was to at least move that deeply nested for-loop into its own method and maybe reduce its nesting a bit. And also see if we can use anything else than a for-loop. But that can also be done later

Ladsgroup:
> I like it.
> I think my original goal was to at least move that deeply nested for-loop into its own method and maybe reduce its nesting a bit. And also see if we can use anything else than a for-loop. But that can also be done later

oh I can do that too. I'll make a follow up PR

PR #192, opened 2021-02-10
Rewrite the existing SauceLabs/Nightwatch Browser Test to Cypress and…

bereket-WMDE:
Bug: [T273775](https://phabricator.wikimedia.org/T273775)


micgro42 (changes requested):
I think there is a bit more here that needs cleaning up:
* the old e2e tests and directory still exist in the `tests/` directory, the package.json still has the commands to run those tests. It also still has dependencies for nightwatch and saucelabs, those need to be uninstalled in a way that also removes them from the package-lock.json

bereket-WMDE:
> I think there is a bit more here that needs cleaning up:
> 
>     * the old e2e tests and directory still exist in the `tests/` directory, the package.json still has the commands to run those tests. It also still has dependencies for nightwatch and saucelabs, those need to be uninstalled in a way that also removes them from the package-lock.json

@micgro42 updated. I also deleted some more unnecessary stuff and deleted the sauclabs tokens from github secrets

micgro42:
There is still a nightwatch.conf.js. But after that is removed, then I think this is good to go

bereket-WMDE:
> There is still a nightwatch.conf.js. But after that is removed, then I think this is good to go

deleted

micgro42:
I noticed `tests_output/QueryBuilder.test.xml` just now. It would seem that was added in error, or was that intentional?

bereket-WMDE:
> I noticed `tests_output/QueryBuilder.test.xml` just now. It would seem that was added in error, or was that intentional?

Yes, good catch. It was added in Error. I removed it now

micgro42 (approved):
\o/

PR #193, opened 2021-02-10
Group reference patterns into one

Ladsgroup:
This way, the resulting query when there is a union and reference filter
wouldn't be wrong anymore.

Bug: T274101


micgro42:
I think if this is rebased on the top of master, then the browser tests should be fine and it should be good to go.

Ladsgroup:
> I think if this is rebased on the top of master, then the browser tests should be fine and it should be good to go.

Done. Thanks.

micgro42 (approved):

PR #194, opened 2021-02-10
fix: debounce searching for options on typing

micgro42:
The debounce in the repository wasn't working as it was recreated on every call.

Unfortunately, lodash.debounce does not properly work with methods that actually return something as it seemingly always returns undefined on first invocation.

Therefore, the debounce is moved to the EntityLookup, and the update to the state happens inside the debounced method.

Bug: T273868

### DIFF
--- a/src/components/EntityLookup.vue
+++ b/src/components/EntityLookup.vue
@@ -19,6 +19,7 @@
 <script lang="ts">
 import { MenuItem } from '@wmde/wikit-vue-components/dist/components/MenuItem';
 import Vue, { PropType } from 'vue';
+import debounce from 'lodash/debounce';
 
 import { Lookup } from '@wmde/wikit-vue-components';
 import SearchResult from '@/data-access/SearchResult';
@@ -36,6 +37,7 @@ export default Vue.extend( {
 			search: '',
 			searchResults: [] as MenuItem[],
 			topItemIndex: 1,
+			debouncedUpdateMenuItems: null as Function | null,
 		};
 	},
 	methods: {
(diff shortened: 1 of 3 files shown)


Ladsgroup:
I'm not 100% sure that's the problem. I don't know much about debounce and how it works (I mostly know it from https://css-tricks.com/debouncing-throttling-explained-examples/) and my biggest issue is that yes it currently doesn't work properly but if you set trailing to false, it does work correctly:
image

I don't know why but setting trailing to false fixes the issue and works just fine.

micgro42:
> I'm not 100% sure that's the problem. I don't know much about debounce and how it works (I mostly know it from [css-tricks.com/debouncing-throttling-explained-examples](https://css-tricks.com/debouncing-throttling-explained-examples/)) and my biggest issue is that yes it currently doesn't work properly but if you set trailing to false, it does work correctly:
> image
> 
> I don't know why but setting trailing to false fixes the issue and works just fine.

But non of those requests are at least 300 ms apart :slightly_smiling_face: 

Also, using only `leading: true` gives us the wrong results, as your link illustrates. If I quickly type the keys "postal", then I would like to see the search results for "postal", not the search results for "p" :upside_down_face:

bereket-WMDE (commented):
tested it locally with 
`docker-compose --env-file .env -f docker-compose.yml -f docker-compose.e2e.yml up e2e`
and its failing. maybe I'm missing something ? I guess we are using different versions of cypress to test locally and on github action

micgro42:
> tested it locally with
> `docker-compose --env-file .env -f docker-compose.yml -f docker-compose.e2e.yml up e2e`
> and its failing. maybe I'm missing something ? I guess we are using different versions of cypress to test locally and on github action

I tried it locally 5 times just now and each test passed. :thinking: 

What error are you seeing? And how does the video look?

bereket-WMDE:
> > tested it locally with
> > `docker-compose --env-file .env -f docker-compose.yml -f docker-compose.e2e.yml up e2e`
> > and its failing. maybe I'm missing something ? I guess we are using different versions of cypress to test locally and on github action
> 
> I tried it locally 5 times just now and each test passed. thinking
> 
> What error are you seeing? And how does the video look?

its working now. I checked out your branch to test it locally but somehow it doesn't have your commit from this PR. 
very strange. sorry, it was my mistake

bereket-WMDE (approved):

PR #195, opened 2021-02-12
Start "Get a shareable link"

Ladsgroup:
Bug: T272887

### DIFF
--- a/cypress/integration/error-handling.spec.js
+++ b/cypress/integration/error-handling.spec.js
@@ -11,9 +11,10 @@ function wikibaseApiRequest( query ) {
 describe( 'Test error handling of the Query Building', () => {
 	it( 'Tests whether the Query builder handles user input error correctly.', () => {
 		cy.visit( '/' );
+		const runQueryButtonSelector = '.querybuilder__run .wikit-Button--progressive';
 
 		// Run query without any input and assert if any error message is displayed
-		cy.get( '.querybuilder__run .wikit-Button' ).click();
+		cy.get( runQueryButtonSelector ).click();
 		cy.get( '.querybuilder__result__errors' ).should( 'be.visible' );
 
 		cy.intercept(
@@ -38,7 +39,7 @@ describe( 'Test error handling of the Query Building', () => {
 		 * Run the query only with a property and no value. And assert wheather
 		 * an error message is displayed on the value input.
 		 */
-		cy.get( '.querybuilder__run .wikit-Button' ).click();
+		cy.get( runQueryButtonSelector ).click();
 		cy.get( '.wikit.wikit-ValidationMessage.wikit-ValidationMessage--error' ).should( 'be.visible' );
 
 		// Add some value to the value input component
@@ -52,7 +53,7 @@ describe( 'Test error handling of the Query Building', () => {
 
 		// Expand the query by clicking 'add condition' and run the query as is.
 		cy.get( '.querybuilder__add-condition button' ).click();
-		cy.get( '.querybuilder__run .wikit-Button' ).click();
+		cy.get( runQueryButtonSelector ).click();
 
 		/**
 		 * there shouldn't be an error on the first block.
@@ -69,7 +70,7 @@ describe( 'Test error handling of the Query Building', () => {
 		cy.get( '.query-condition:nth(1) .delete-condition-button' ).click();
 
 		// Run query
-		cy.get( '.querybuilder__run .wikit-Button' ).click();
+		cy.get( runQueryButtonSelector ).click();
 
 		// Assert the resulting sparql query and make sure there are no errors generated.
 		cy.get( '.querybuilder__result__iframe' ).then( ( element ) => {
--- a/src/components/SharableLink.vue
+++ b/src/components/SharableLink.vue
@@ -1,0 +1,78 @@
+<template>
+	<span class="querybuilder__sharable-link">
+		{{ $i18n('query-builder-sharable-link-text') }}
+		<Popover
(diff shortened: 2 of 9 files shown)


cypress/integration/error-handling.spec.js
```diff
@@ -69,7 +69,7 @@ describe( 'Test error handling of the Query Building', () => {
 		cy.get( '.query-condition:nth(1) .delete-condition-button' ).click();
 
 		// Run query
-		cy.get( '.querybuilder__run .wikit-Button' ).click();
+		cy.get( '.querybuilder__run .wikit-Button--progressive' ).click();
```
bereket-WMDE: we should make this a variable. so we don't have to change names everywhere.
something like this at the top:

`const runQueryButton = cy.get('.querybuilder__run .wikit-Button--progressive')
`

and then use this variable everywhere needed.

`runQueryButton.click();`

src/components/SharableLink.vue
```diff
@@ -0,0 +1,77 @@
+<template>
+	<span class="querybuilder__sharable-link">
+		{{ $i18n('query-builder-sharable-link-text') }}
+		<Popover
+			:reactToHover="false"
+			position="end"
+		>
```
bereket-WMDE, line 7: There's a separate ticket to add the Popover component into the QB.
see [here](https://phabricator.wikimedia.org/T275542). 
If the popover is supposed to react with hover (eventually), we might forget about it by then. I'd say we don't need this here now. but should be done in the dedicated ticket. 

I know it says it in the ticket. but there's a ticket dedicated for dealing with popovers in the QB

src/components/SharableLink.vue
```diff
@@ -0,0 +1,77 @@
...
+		async copyTextToClipboard(): Promise<void> {
+			const querySerializer = new QuerySerializer();
+			const serializedQuery = querySerializer.serialize( this.$store.state );
+			const current = new URL( window.location.href );
+			current.searchParams.set( 'query', serializedQuery );
+			this.href = current.href;
+			try {
+				await navigator.clipboard.writeText( current.href );
+			} catch ( err ) {
+				// pass
```
bereket-WMDE: what does 'pass' mean. is something supposed to happen here ?

bereket-WMDE (changes requested):

bereket-WMDE:
linkicon

The icon , inside the button, doesn't seem to be vertically centered.

cypress/integration/error-handling.spec.js
Ladsgroup: Done

src/components/SharableLink.vue
Ladsgroup: the component is not going react to hover (given that it needs to generate short url) like never. I know there's a dedicated ticket for using wikit popover but it's done and I'm doing both at the same time, specially since the html native tolltip is not easy to implement for this usecase.

src/components/SharableLink.vue
Ladsgroup: good point. let me just remove it.

cypress/integration/error-handling.spec.js
Ladsgroup: It errors with:
```
error: Do not assign the return value of a Cypress command (cypress/no-assigning-return-values) at cypress/integration/error-handling.spec.js:14:3:
```
I make the text a variable.

src/components/SharableLink.vue
Ladsgroup: Removing the catch makes some tests to warn, I changed it to a better comment why it should ignore failure.

Ladsgroup:
> linkicon
> 
> The icon , inside the button, doesn't seem to be vertically centered.

That is coming from the wikit. I think it should be fixed there (file a bug, etc.)

cypress/integration/error-handling.spec.js
bereket-WMDE: thats a pity. actually, this error was fixed in version 2.11.0 of cypress. 
however, we are using cypress-io, and I'm afraid this issue is not resolved there. 
but making the text variable is sufficient here

src/components/SharableLink.vue
bereket-WMDE: makes sense. I was being pedantic anyways :)

bereket-WMDE:
> That is coming from the wikit. I think it should be fixed there (file a bug, etc.)

I will check in with Sarai about this

bereket-WMDE:
looks good to me

bereket-WMDE (approved):

src/components/SharableLink.vue
```diff
@@ -0,0 +1,78 @@
...
+			position="end"
+		>
+			<template v-slot:target>
+				<Button
+					variant="normal"
+					type="neutral"
+					iconOnly
+					@click.native="copyTextToClipboard"
+					aria-label="Sharable link">
+					<Icon type="link" size="medium" />
```
sai-san, line 15: the default size of icons used in iconOnly buttons should be "large"

Ladsgroup: Even with large, it doesn't fix the center alignment
image

sai-san: didn't mean to imply that that was the only issue. It's just what I noticed right away (҂◡_◡) 
I also noticed that for some reason the button changes if the text accompanying it is modified, so there might be some inheritance issues there (maybe related to the line-height)? Or maybe we need to specify vertical-alignment (can't see that applied). In any case, it looks like we need a separate ticket

PR #196, opened 2021-02-15
Value_type_not_detected_in_the_visual_query_after_using_regardless_of…

bereket-WMDE:
[Bug: T273052](https://phabricator.wikimedia.org/T273052)

### DIFF
--- a/src/components/EntityLookup.vue
+++ b/src/components/EntityLookup.vue
@@ -78,6 +78,11 @@ export default Vue.extend( {
 		},
 	},
 	watch: {
+		disabled( isDisabled: boolean ): void {
+			if ( isDisabled ) {
+				this.search = '';
+			}
+		},
 		search( newSearchString: string ): void {
 			this.topItemIndex = 0;
 			if ( !newSearchString ) {
--- a/src/components/QueryCondition.vue
+++ b/src/components/QueryCondition.vue
@@ -117,6 +117,7 @@ export default Vue.extend( {
 				if ( selectedPropertyValueRelation === PropertyValueRelation.Regardless ) {
 					this.conditionValue = null;
 				}
+
 				this.$store.dispatch(
 					'updatePropertyValueRelation',
 					{ propertyValueRelation: selectedPropertyValueRelation, conditionIndex: this.conditionIndex },


micgro42:
Could you say more about why you chose that particular technical solution?
I would like to understand your motivation for choosing it better in order to be able to review it.

bereket-WMDE:
> Could you say more about why you chose that particular technical solution?
> I would like to understand your motivation for choosing it better in order to be able to review it.

it definitely was not my first choice of solution. I first tried watching for props. I added a prop in the QueryCondition.vue file and set the value to true when the property value relation changes: i.e
`set( selectedPropertyValueRelation: PropertyValueRelation ): void {
				if ( selectedPropertyValueRelation === PropertyValueRelation.Regardless ) {
					this.isRelationRegardless = true
				}
				this.$store.dispatch(
					'updatePropertyValueRelation',
					{ propertyValueRelation: selectedPropertyValueRelation, conditionIndex: this.conditionIndex },
				);
			},`
			
and added a watch on EntityLookup.vue file to watch for changes for that prop. However, that didn't work. somehow the value change is not reflected. I then tried to just emit an event and listen to that event on EntityLookup.vue file. however, the event is not propagated. 

hence why i decided to use $root to emit the event.

P.S. another solution I could think of is, watching the store in the EntitiyLookup.vue file. i.e watching `this.$store.getters.PropertyValueRelation.value` ?? maybe. Its just a thought

micgro42 (changes requested):
Mh, the current solution isn't working because 
1) it affects both the PropertyLookup and the ItemValueLookup and
2) it affects every ItemValueLookup (and every PropertyLookup) in every condition

What we probably need is data flowing down via props from the QueryCondition to the ValueInput components.
That being said - we already have this. We probably could just set this search string to empty when the `disabled` prop is flipped to `true`. Because we never want there to be seemingly unchangeable data in a disabled input field.

bereket-WMDE:
> Mh, the current solution isn't working because
> 
>     1. it affects both the PropertyLookup and the ItemValueLookup and
> 
>     2. it affects every ItemValueLookup (and every PropertyLookup) in every condition
> 
> 
> What we probably need is data flowing down via props from the QueryCondition to the ValueInput components.
> That being said - we already have this. We probably could just set this search string to empty when the `disabled` prop is flipped to `true`. Because we never want there to be seemingly unchangeable data in a disabled input field.

@micgro42 
thank you very much for the explanation. the solution was very obvious and i went about it in a more convoluted direction.
I updated the PR

src/components/QueryCondition.vue
```diff
@@ -117,9 +117,6 @@ export default Vue.extend( {
 				return this.$store.getters.propertyValueRelation( this.conditionIndex );
 			},
 			set( selectedPropertyValueRelation: PropertyValueRelation ): void {
-				if ( selectedPropertyValueRelation === PropertyValueRelation.Regardless ) {
-					this.textInputValue = '';
```
micgro42: Why did you change this?

bereket-WMDE: that was not working in the first place. It was throwing an error on the console.
`textInputValue` is not a string (its an object). and I suspect it was put here assuming it will do what this PR is trying to do. 
since its not doing anything, I don't see the purpose of that line of code.

bereket-WMDE: I see you may have addressed the issue [here](https://github.com/wmde/query-builder/pull/198/files#diff-216bcf52888451e05bc28ab7fbccc04d1bb8d4181f7bddae00f1ea8fd9f650fdR122)

bereket-WMDE: I reverted it. I will merged this one after you merge [198](https://github.com/wmde/query-builder/pull/198)

src/components/QueryCondition.vue
```diff
@@ -120,6 +120,7 @@ export default Vue.extend( {
 				if ( selectedPropertyValueRelation === PropertyValueRelation.Regardless ) {
 					this.textInputValue = '';
 				}
+
```
micgro42, line 120: stray newline

src/components/EntityLookup.vue
```diff
@@ -78,6 +78,9 @@ export default Vue.extend( {
 		},
 	},
 	watch: {
+		disabled( isDisabled: boolean ): void {
+			this.search = isDisabled ? '' : this.search;
```
micgro42: I'm not sure what were gaining by assigning `this.search` to itself.

What do you think about the following?

```suggestion
			if ( isDisabled ) {
			  this.search = '';
			}
```

bereket-WMDE: much cleaner

src/components/QueryCondition.vue
bereket-WMDE: rebased with master and now fixed.

micgro42 (approved):

PR #197, opened 2021-02-15
Implement UX improvements

Ladsgroup:
As outlined in the ticket, this PR implements these improvements from
A-F

Bug: T274283

### DIFF
--- a/public/i18n/qqq.json
+++ b/public/i18n/qqq.json
@@ -7,6 +7,8 @@
 	},
 	"query-builder-heading": "Title of the page",
 	"query-builder-intro-text": "Intro text describing what query builder is.",
+	"query-builder-query-heading": "Heading of the section for building the query",
+	"query-builder-settings-heading": "Heading of the section for query settings",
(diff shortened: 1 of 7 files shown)


public/i18n/qqq.json
```diff
@@ -7,6 +7,8 @@
 	},
 	"query-builder-heading": "Title of the page",
 	"query-builder-intro-text": "Intro text describing what query builder is.",
+	"query-builder-query": "Name of the section for building the query",
```
micgro42: "Name" is a bit unspecific with respect to how it is used. How about "Heading"? Maybe we could also include that in the message-key?

micgro42 (commented):
Looks good, just a minor phrasing issue about the messages

public/i18n/qqq.json
Ladsgroup: Yeah, good point. Added to both message documentation and message name

micgro42 (approved):

PR #198, opened 2021-02-15
refactor: adjust obsolete names and types in QueryCondition

micgro42:
Since we are no longer only using a TextInput as the input for the
value, the names should also be more general.

Also, this might fix the error message in the console when switching the
relation to "regardless of value".

### DIFF
--- a/src/components/QueryCondition.vue
+++ b/src/components/QueryCondition.vue
@@ -26,14 +26,14 @@
 			<div>
 				<ValueInput
 					class="query-condition__value-input"
-					:disabled="isTextInputDisabled()"
-					v-model="textInputValue"
+					:disabled="isValueInputDisabled()"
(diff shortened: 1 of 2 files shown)


bereket-WMDE (approved):

src/components/QueryCondition.vue
```diff
@@ -118,7 +119,7 @@ export default Vue.extend( {
 			},
 			set( selectedPropertyValueRelation: PropertyValueRelation ): void {
 				if ( selectedPropertyValueRelation === PropertyValueRelation.Regardless ) {
-					this.textInputValue = '';
+					this.conditionValue = null;
```
bereket-WMDE, line 122: nice. this DOES fix the problem!

bereket-WMDE (approved):

PR #199, opened 2021-02-15
fix: show subclass checkbox for item properties when reconstructing

micgro42:
This resolves the issue that the subclass-checkbox was not shown for properties with datatype `wikibase-item` when reconstructing the Query Builder from a link.

Bug: T274659


bereket-WMDE (approved):
Looks good!
tested and works.

PR #203, opened 2021-02-17
expand_existing_browser_tests

bereket-WMDE:
[Bug: T273777](https://phabricator.wikimedia.org/T273777)

wrote the tests according to the BDDs in the ticket. It can be easily expanded if we want to cover more stuff.
to run the test locally simply run:

`CYPRESS_BASE_URL=http://localhost:8080 CYPRESS_API_URL=https://www.wikidata.org/w/api.php npx cypress open`

It will open an interactive browser window so you can see exactly whats happening.


bereket-WMDE:
wrote the tests according to the BDDs in the ticket. It can be easily expanded if we want to cover more stuff.
to run the test locally simply run: 
`CYPRESS_BASE_URL=http://localhost:8080 CYPRESS_API_URL=https://www.wikidata.org/w/api.php npx cypress open`

It will open an interactive browser window so you can see exactly whats happening.

cypress/integration/component-interaction.spec.js
```diff
@@ -0,0 +1,144 @@
...
+
+        cy.get( '.query-condition__value-input .wikit-Input' )
+            .type( 'house cat' )
+            .wait( '@houseCatRequest' );
+        cy.get( '.query-condition__value-input .wikit-OptionsMenu__item' )
+            .first()
+            .click();
+
+        // clear out values from specified input components
+        cy.get( '.query-condition__property-lookup .wikit-Input' ).clear();
```
micgro42: What is the motivation behind clearing the inputs as opposed to adding another condition?
The original plan was to have 4 conditions (string, item, external-id, limited support) when running the query and asserting the SPARQL.

While it is totally fine to deviate from that, the reasoning why that is done should be communicated somewhere :)

micgro42 (changes requested):
Already looking great, but there are some things left to.

This is currently mixing tabs and spaces, we should adjust eslint to lint the files here as well.

Also, I think we can remove the two existing tests. I kind of had expected them to morph into the new ones. I don't think they test anything useful anymore once we merge these here.

cypress/integration/component-interaction.spec.js
bereket-WMDE: good feedbacks :)

- I added a cypress lint plugin and added some rules for linting.
- I agree with the other two tests. they are kinda redundant at this point. and the two new tests are pretty comprehensive and cover all the scenarios the old tests were doing. I deleted them.
- I'm trying to cover more 'realistic' user interaction scenarios. all the 4 conditions ((string, item, external-id, limited support) ) have been covered in different scenarios.

package.json
```diff
@@ -8,6 +8,7 @@
   },
   "scripts": {
     "cypress": "CYPRESS_BASE_URL=${CYPRESS_BASE_URL:-http://localhost:8080} cypress run",
+    "test:lint-cypress": "eslint --fix cypress/integration/**",
```
micgro42:
```suggestion
    "test:lint-cypress": "eslint --no-fix --max-warnings 0 cypress/integration/**",
```

Mh, this should be similar to `test:lint` with the `--no-fix` and `--max-warning 0` parameters. Though having a fix:cypress command might make sense indeed.

At some point we probably want to use `vue-cli-service lint` instead of eslint directly and maybe also lint the entire `cypress/` directory and not just the `integration/` subdirectory, but I'm still getting an error there that need some more thought. We can do that later.

cypress/integration/component-interaction.spec.js
micgro42: Not sure I follow entirely, especially because I didn't see the string datatype, but this is already much better than what we had before.
Maybe somebody else can add more to the browser test so that we spread the knowledge of writing Cypress browser tests around.

micgro42 (changes requested):
Almost ready, but the command to execute eslint needs a bit more work

cypress/integration/component-interaction.spec.js
bereket-WMDE: I agree. we should write more browser tests (for upcoming components) to spread the knowledge. 
I was actually talking about the `@hasPetRequest`. its a string search.... AFAIK,

maybe we should make browser testing a part of development (like unit testing). so people can add more tests, whenever a UI task is involved

package.json
bereket-WMDE: - good suggestion about the `--no-fix` thing
- yea, I tested the whole `cypress/` directory. There are some errors that need more investigation. Its mostly in the `cypress/plugins` directory.

micgro42 (approved):

PR #205, opened 2021-02-18
fix: complete the signature of the i18n plugin

micgro42:
The i18n allows for providing parameters to the message. These are
usually strings, but could also be numbers, therefore choosing `unknown`
as the type.

Providing the correct signature resolves errors in the IDE that
complained about multiple arguments being provided to a method that only
expected one.

### DIFF
--- a/src/shims-vue.d.ts
+++ b/src/shims-vue.d.ts
@@ -3,7 +3,7 @@ declare module '*.vue' {
 
 	declare module 'vue/types/vue' {
 		interface Vue {
-			$i18n: ( msg: string ) => string;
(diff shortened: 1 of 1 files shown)


Ladsgroup:
Maybe `any` instead of `unknown` but I don't know good enough to say which one is better.

micgro42:
In this specific case it makes no difference. In general, `any` shouldn't be used as it disables type checking.
`any` can be ok in tests.

Ladsgroup:
Thanks for the explanation!

Ladsgroup (approved):

PR #206, opened 2021-02-19
chore: better cypress linting

micgro42:
This keeps the linting config for cypress confined to that directory and
also lints the code in the entire cypress directory.

It also adds a fix command.


cypress/plugins/index.js
```diff
@@ -1,4 +1,4 @@
-/// <reference types="cypress" />
+// / <reference types="cypress" />
```
bereket-WMDE, line 1: this one was bugging me the whole time :)

bereket-WMDE (approved):

PR #207, opened 2021-02-19
refactor: simplify method of collecting unions

micgro42:
It should now be hopefully more readable.

### DIFF
--- a/src/sparql/QueryObjectBuilder.ts
+++ b/src/sparql/QueryObjectBuilder.ts
@@ -1,7 +1,6 @@
 import rdfNamespaces from '@/sparql/rdfNamespaces';
 import QueryRepresentation, { Condition } from '@/sparql/QueryRepresentation';
 import { Pattern, SelectQuery } from 'sparqljs';
-import ConditionRelation from '@/data-model/ConditionRelation';
 import PatternBuilder from '@/sparql/PatternBuilder';
 
 export default class QueryObjectBuilder {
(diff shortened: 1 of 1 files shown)


src/sparql/QueryObjectBuilder.ts
```diff
@@ -67,36 +66,41 @@ export default class QueryObjectBuilder {
...
+				this.attachToRootWithAnd( rootNode, condition );
+				return;
 			}
+			this.attachToPreviousConditionWithOr( rootNode, condition );
+		} );
+
+		return rootNode;
+	}
+
+	private attachToPreviousConditionWithOr( rootNode: ( Condition | Condition[] )[], condition: Condition ): void {
```
Ladsgroup: Nitpick. Can you define "RootNode" type? as ( Condition | Condition[] )[]. This has been repeated a lot and I also like documentation on what root node is and stuff like that too. Of course, this can happen in a follow up.

Ladsgroup: Maybe even as a class having methods like "node.lastElementIsSingleCondition()"

micgro42: > Nitpick. Can you define "RootNode" type? as ( Condition | Condition[] )[]. This has been repeated a lot and I also like documentation on what root node is and stuff like that too. Of course, this can happen in a follow up.

Good point, will do!

> Maybe even as a class having methods like "node.lastElementIsSingleCondition()"

Yeah, I thought about extracting classes here, but it felt like overkill because we are not going to add any more complexity to it in this hike (i.e. no deeper levels of nesting conditions).
And we have no idea if we will ever add those additional nesting levels or when that will happen. So, to me it would seem rather [yagni](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it) for now.

Ladsgroup (approved):

PR #211, opened 2021-02-25
[PoC] Start using scss mixins to roll up text styles

Ladsgroup:
Reduce logic duplication.


bereket-WMDE:
@Ladsgroup I like the idea

Ladsgroup:
I'll do it once we are in the gerritland.

PR #213, opened 2021-03-02
Make a mapping of subclass type

Ladsgroup:
So for example, when checking for monuments in Germany, it would make a
correct query for it.


Ladsgroup:
Try these two queries on this:
```
http://localhost:8080/?query=%7B%22conditions%22%3A%5B%7B%22propertyId%22%3A%22P171%22%2C%22propertyDataType%22%3A%22wikibase-item%22%2C%22propertyValueRelation%22%3A%22matching%22%2C%22referenceRelation%22%3A%22regardless%22%2C%22value%22%3A%22Q42602%22%2C%22subclasses%22%3Atrue%2C%22conditionRelation%22%3Anull%2C%22negate%22%3Afalse%7D%2C%7B%22propertyId%22%3A%22P141%22%2C%22propertyDataType%22%3A%22wikibase-item%22%2C%22propertyValueRelation%22%3A%22matching%22%2C%22referenceRelation%22%3A%22regardless%22%2C%22value%22%3A%22Q211005%22%2C%22subclasses%22%3Atrue%2C%22conditionRelation%22%3A%22and%22%2C%22negate%22%3Afalse%7D%5D%2C%22limit%22%3A100%2C%22useLimit%22%3Atrue%2C%22omitLabels%22%3Afalse%7D

http://localhost:8080/?query=%7B%22conditions%22%3A%5B%7B%22propertyId%22%3A%22P31%22%2C%22propertyDataType%22%3A%22wikibase-item%22%2C%22propertyValueRelation%22%3A%22matching%22%2C%22referenceRelation%22%3A%22regardless%22%2C%22value%22%3A%22Q2977%22%2C%22subclasses%22%3Atrue%2C%22conditionRelation%22%3Anull%2C%22negate%22%3Afalse%7D%2C%7B%22propertyId%22%3A%22P131%22%2C%22propertyDataType%22%3A%22wikibase-item%22%2C%22propertyValueRelation%22%3A%22matching%22%2C%22referenceRelation%22%3A%22regardless%22%2C%22value%22%3A%22Q1198%22%2C%22subclasses%22%3Atrue%2C%22conditionRelation%22%3A%22and%22%2C%22negate%22%3Afalse%7D%5D%2C%22limit%22%3A100%2C%22useLimit%22%3Atrue%2C%22omitLabels%22%3Afalse%7D
```

bereket-WMDE (commented):
I like the idea. In case you/ we decide to move forward with this, don't forget to remove the `VUE_APP_SUBCLASS_PROPERTY` from the following files.

- `.jest/setEnvVars.js`
- `tests/unit/sparql/buildQuery.spec.ts`

.jest/setEnvVars.js
```diff
@@ -1 +1 @@
-process.env.VUE_APP_SUBCLASS_PROPERTY = 'P279';
+process.env.VUE_APP_SUBCLASS_PROPERTY_MAP = '{"default": "P279", "P171": "P171", "P131":"P131"}';
```
guergana: I must disagree on this one.. :sweat_smile: 

Environment variables are set on this file so you can later invoke them like this:

`process.env.VUE_APP_SUBCLASS_PROPERTY;` as it was done before the change in https://github.com/wmde/query-builder/pull/213/files#diff-33a89579bad0cce26e49f525b9e3d5320d953ef16e2f40f1dca270c5c0917fb4R265 .

If you assign it a value in the test like this `process.env.VUE_APP_SUBCLASS_PROPERTY_MAP = '{"P666":"P66"}';` declaring it here in `setEnvVars` is not needed. In this case you need to delete this file and also remove this line `setupFiles: [ '<rootDir>/.jest/setEnvVars.js' ],` from the file: `/.jest.config.js`

I personally prefer having them configured in the `setEnvVars` file so it behaves more like production and we have all the enviroment variables using in the unit tests in one place.

In any case this file in this PR is not being used so it generates confusion. It should either be removed or used properly. :)

guergana (changes requested):

.jest/setEnvVars.js
Ladsgroup: The test being changed is a unit test and those by definition should not depend on the global state. I would rather keep this file just to make sure tests work for example if the code checks if this env variable exists and fails to do anything if not. This would make sure it runs.

guergana: Try removing it first. It works locally but I remember having some issues once pushed... if it is not being used it should be deleted IMHO.

micgro42: I tried it out and these files are indeed no longer needed. I took the liberty to remove them by amending this PR as it is prone to cause conflicts with other work on SPARQL generation. @guergana could you have another look?

guergana: Sure. Looks good and also the query works. Will approve

guergana (approved):

.jest/setEnvVars.js
micgro42: Thank you :pray:

PR #216, opened 2021-03-10
Start InfoTooltip component

Ladsgroup:
Bug: T275542


micgro42:
I know what the test failure is. Will fix it in the Icon component in storybook

guergana:
Looks great so far. Is the blue border that appears on focus supposed to be part of the implementation? I am guessing it is but the margin/padding doesn't look right. 

image

Ladsgroup:
Nice catch. Fixed.

src/components/InfoTooltip.vue
```diff
@@ -0,0 +1,62 @@
...
+</script>
+
+<style lang="scss">
+.querybuilder__Tooltip {
+	& .wikit-Popover__target .wikit.wikit-Button--iconOnly.wikit-Button--neutral {
+		padding-block: 0;
+		padding-inline-start: 0;
+		line-height: 1;
+
+		&:focus {
```
guergana: There is still a jumpy thing with a border that looks strange. If you click on the icon and then outside you can briefly see a jump and a black border. 

I tried with:
``` 
& .wikit-Popover__target .wikit.wikit-Button--iconOnly.wikit-Button--neutral {
		padding-block: 0;
		padding-inline-start: 0;
		line-height: 1;
		border: 0;

		&:focus {
			box-shadow: none;
		}

	}
```

and it works. 

I am not sure about the margin still, because there is a gray area around the icon and you can still see it is aligned to the start. I can see it in the css above, why did you decide to align it like that?

guergana (changes requested):

src/components/InfoTooltip.vue
Ladsgroup: It's because it pushes the input fields down. That's why I removed the padding.

Ladsgroup: I made some changes. I think it's much better now.

guergana:
Pasting here @SaiSan-WMDE 's comment from the chat. 

>  "focusable" elements should display focus state. It's a matter of centering the icon and make the border even in all sides. I'm not sure if we need to modify the span with class="wikit wikit-Icon wikit-Icon--medium wikit-Icon--base"in WiKit (make it 20/20px by default) or if we just need to make the adjustments in the Query builder 

Also please the conflicts. :grimacing: then it should be good. Sorry for being the painnnn.

Ladsgroup:
Fixed the rebase and centering of the icon. I think it's in an acceptable shape now.

guergana:
Sorry, it looks the same as before removing the focus border. It is still not even on all sides. (slightly bigger margin at the bottom) 

image

Also the checks didn't get executed in the last push. Why is that??

Ladsgroup:
> Sorry, it looks the same as before removing the focus border. It is still not even on all sides. (slightly bigger margin at the bottom)
> 

Sorry It seems I didn't push properly. Is it better now?

sai-san:
I'd like to check this, but I'm unable to find the Netlify link anywhere :-/

guergana:
I approve it but please dont merge until Sarai has given her veredict. 

> I'd like to check this, but I'm unable to find the Netlify link anywhere :-/

I also didn't find it but you can get the code locally and see at `localhost:8080`

guergana (approved):

sai-san:
Some comments after checking the current status with Michael's help. Disclaimer: I couldn't interact with the QB myself. The border now looks good enough, but we noticed that:

- The copy of the Value label popover seems to be missing (see [Phab ticket](https://phabricator.wikimedia.org/T275542))
- Popovers should be hidden when users either press `enter`again, `esc` or if they `tab` away from the trigger element (button) to focus on the next focusable element on the page.
- The current aria-label applied to the popover target (`aria-label="Tooltip"`) is not very helpful. I would suggest replacing it by something like "open tooltip with more information".

Ladsgroup:
> Some comments after checking the current status with Michael's help. Disclaimer: I couldn't interact with the QB myself. The border now looks good enough, but we noticed that:
> 
>     * The copy of the Value label popover seems to be missing (see [Phab ticket](https://phabricator.wikimedia.org/T275542))

It doesn't seem missing to me.
image

> 
>     * Popovers should be hidden when users either press `enter`again, `esc` or if they `tab` away from the trigger element (button) to focus on the next focusable element on the page.

This is supposed to be handled in wikit, not here. Can we have a ticket for it?
> 
>     * The current aria-label applied to the popover target (`aria-label="Tooltip"`) is not very helpful. I would suggest replacing it by something like "open tooltip with more information".

Done.

sai-san:
> > ```
> > * Popovers should be hidden when users either press `enter`again, `esc` or if they `tab` away from the trigger element (button) to focus on the next focusable element on the page.
> > ```
> 
> This is supposed to be handled in wikit, not here. Can we have a ticket for it?

That's right. I'm not really sure how come it's not handled in WiKit. A ticket to create the missing events is now available in the SQB board: [T277738](https://phabricator.wikimedia.org/T277738)

src/components/ItemValueLookup.vue
```diff
@@ -5,6 +5,7 @@
 		:error="error"
 		:searchForMenuItems="searchForItems"
 		:label="$i18n('query-builder-input-value-label')"
+		:tooltip="$i18n('query-builder-item-value-tooltip')"
```
micgro42: This is the i18n message that doesn't exist. :)

But that being said, it probably shouldn't be created, but we should inject that message in the ValueInput component. Because, as I understand it, we do not have different messages for the different datatypes, right? What do you think?

micgro42 (changes requested):

src/components/ItemValueLookup.vue
Ladsgroup: oh yeah, I initially thought only item value needs it but during the change, I forgot to fix this one.

micgro42 (approved):

PR #218, opened 2021-03-11
Adjust Value interface in RootStore by adding QuantityValue type to it

guergana:


src/store/RootState.ts
```diff
@@ -18,7 +18,13 @@ export interface ItemValue {
 
 export type StringValue = string;
 
-export type Value = ItemValue | StringValue | null;
+export interface QuantityValue {
+	value: number;
+	precision?: number;
+	unit: ItemValue;
```
micgro42: The unit should either be also optional or it should be nullable.

micgro42 (changes requested):
Almost there! What is missing is the ability for this type to represent a value without unit, e.g. a population 10000.

src/store/RootState.ts
guergana: adressed

micgro42 (approved):
Looks good, thank you :)

PR #219, opened 2021-03-11
Add UnitValue type to QueryRepresentation

guergana:
…unt for unit values

### DIFF
--- a/src/sparql/TripleBuilder.ts
+++ b/src/sparql/TripleBuilder.ts
@@ -3,6 +3,7 @@ import { IriTerm, PropertyPath, Term, Triple } from 'sparqljs';
 import PropertyValueRelation from '@/data-model/PropertyValueRelation';
 import rdfNamespaces from '@/sparql/rdfNamespaces';
 import ReferenceRelation from '@/data-model/ReferenceRelation';
+import UnitValue from '@/data-model/UnitValue';
 
 export default class TripleBuilder {
 	public buildTripleFromQueryCondition( condition: Condition, conditionIndex: number ): Triple {
@@ -19,6 +20,17 @@ export default class TripleBuilder {
 		};
 	}
 
+	private checkValueType( value: string | UnitValue ): string | number {
+		if ( typeof value === 'string' ) {
+			return value;
+		}
+		try {
+			return value.value;
+		} catch ( e ) {
+			throw new Error( 'Unit Values not yet supported!!' );
+		}
+	}
+
 	private buildTripleForObjectItems( condition: Condition ): Term {
 		switch ( condition.propertyValueRelation ) {
 			case ( PropertyValueRelation.NotMatching ):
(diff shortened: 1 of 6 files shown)


src/sparql/TripleBuilder.ts
```diff
@@ -19,6 +20,16 @@ export default class TripleBuilder {
 		};
 	}
 
+	private checkValueType( value: string | UnitValue ): string | number {
+		if ( typeof value === 'string' ) {
+			return value;
+		}
+		if ( 'value' in value ) { // if value is UnitValue
+			return value.value; // TODO: multiply by unit, define how to manage it
```
guergana: Are we doing any unit conversion at some point??

micgro42: Yes, as it was [discussed and documented](https://phabricator.wikimedia.org/T276938), we will have two different queries depending on whether this is a unit or not.

In this specific case here I would suggest to throw an error after the first `if`, instead of returning a number. That is because that number will still lead to a wrong query and so it is better to fail explicitly and loudly.

The proper implementation will be done in the course of [T276938](https://phabricator.wikimedia.org/T276938).

src/sparql/TripleBuilder.ts
```diff
@@ -32,19 +43,20 @@ export default class TripleBuilder {
...
 			default:
 				throw new Error( `unsupported relation: ${condition.propertyValueRelation}` );
 		}
 	}
 
-	private buildTripleForExplicitValue( datatype: string, value: string ): Term {
+	private buildTripleForExplicitValue( datatype: string, value: string | number ): Term {
 		switch ( datatype ) {
 			case 'string':
+			case 'quantity':
```
micgro42: This line is probably causing the browser test failure. We still want to treat `quantity` as a datatype with only limited support in this pull request. We'll build the SPARQL in https://phabricator.wikimedia.org/T276938

src/sparql/TripleBuilder.ts
guergana: Sounds good!

tests/unit/sparql/QueryObjectBuilder.spec.ts
```diff
@@ -67,6 +67,32 @@ describe( 'QueryObjectBuilder', () => {
 		expect( actual ).toStrictEqual( expected );
 	} );
 
+	it( 'with quantity value', () => {
```
guergana, line 70: added this test to pass sonarcloud!!

src/sparql/TripleBuilder.ts
```diff
@@ -19,6 +20,17 @@ export default class TripleBuilder {
...
 
+	private checkValueType( value: string | UnitValue ): string | number {
+		if ( typeof value === 'string' ) {
+			return value;
+		}
+		try {
+			return value.value;
+		} catch ( e ) {
+			throw new Error( 'Unit Values not yet supported!!' );
+		}
```
micgro42:
```suggestion
		throw new Error( 'Unit Values not yet supported!!' );
```

Is there a case where value is not a string but an object that has a value key, and it is not a UnitValue?

tests/unit/sparql/QueryObjectBuilder.spec.ts
micgro42: Looks good to me

src/sparql/TripleBuilder.ts
guergana: I tried this first and it was making the test 'with quantity value' in tests/unit/sparql/QueryObjectBuilder.spec.ts always throw the error. But now it works. :)

PR #220, opened 2021-03-11
ci: run cypress as part of npm test for migration to Jenkins

micgro42:
With this we should no longer need branch hosting to run the browser tests. The Cypress tests will now run as part of `npm test`. This should hopefully allow us to run all the essential testing in the standard Jenkins Node docker container / job that comes with Browsers preinstalled.

Bug: [T277060](https://phabricator.wikimedia.org/T277060)

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
     "name": "The Wikidata team"
   },
   "scripts": {
-    "cypress": "CYPRESS_BASE_URL=${CYPRESS_BASE_URL:-http://localhost:8080} cypress run",
+    "cypress": "CYPRESS_API_URL=${VUE_APP_WIKIBASE_API_URL} CYPRESS_BASE_URL=${CYPRESS_BASE_URL:-http://localhost:8080} cypress run --browser chromium --headless",
     "test:lint-cypress": "vue-cli-service lint --no-fix --max-warnings 0 cypress",
     "serve": "vue-cli-service serve",
     "build": "vue-cli-service build --mode production",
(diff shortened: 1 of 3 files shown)


Ladsgroup:
In general looks good. I just don't know why package-lock.json got changed. We are not installing any new dependencies. Do we?

package.json
```diff
@@ -17,6 +17,7 @@
     "test": "npm-run-all test:* && vue-cli-service build",
     "test:lintcss": "stylelint '**/*.vue'",
     "test:a11y": "jest --testMatch=\"**/tests/a11y/**\"",
+    "test:cy": "env $(grep -v '^#' .env.example | xargs) start-server-and-test serve http-get://localhost:8080 cypress",
```
bereket-WMDE, line 20: Is there a way to consolidate this line with line 10 ? 
AFAIK, they're both running the same cypress tests (the integration tests we wrote). or am I missing something ?
and If not, do we need to specify the `headless` and `browser` options here as well ?

micgro42:
> In general looks good. I just don't know why package-lock.json got changed. We are not installing any new dependencies. Do we?

Yes we are: `"start-server-and-test": "^1.12.0",` it is needed to wait for the dev server to be up and running before starting Cypress.

package.json
micgro42: The command in line 20 is _using_ the command in line 10. I.e., the last `cypress` in line 20 tells it to run `npm run cypress` as soon as it gets a 200 response from the get request to `localhost:8080`.

See the documentation for how `start-server-and-test` works: https://github.com/bahmutov/start-server-and-test#use

bereket-WMDE: thank you!
It looks good to me.

bereket-WMDE (approved):

PR #221, opened 2021-03-11
Accessibility improvement: set focus to and/or-toggle after new condi…

bereket-WMDE:
Bug: [T274273](https://phabricator.wikimedia.org/T274273)

### DIFF
--- a/src/components/QueryBuilder.vue
+++ b/src/components/QueryBuilder.vue
@@ -120,9 +120,12 @@ export default Vue.extend( {
 					.getElementsByClassName( 'querybuilder__condition-wrapper-last' )[ 0 ]
 					.scrollIntoView( { behavior: 'smooth' } );
 
-				const toggle = document.getElementsByClassName( 'querybuilder__condition-relation-toggle' );
-				( toggle[ toggle.length - 1 ] as HTMLElement ).setAttribute( 'tabindex', '0' );
-				( toggle[ toggle.length - 1 ] as HTMLElement ).focus();
+				const toggle = document
+					.querySelectorAll( '.querybuilder__condition-relation-toggle .wikit-ToggleButton' );
+				const toggleElement = toggle[ toggle.length - 1 ] as HTMLElement;
+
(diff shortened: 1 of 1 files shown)


src/components/QueryBuilder.vue
```diff
@@ -119,6 +119,10 @@ export default Vue.extend( {
 				document
 					.getElementsByClassName( 'querybuilder__condition-wrapper-last' )[ 0 ]
 					.scrollIntoView( { behavior: 'smooth' } );
+
+				const toggle = document.getElementsByClassName( 'querybuilder__condition-relation-toggle' );
+				( toggle[ toggle.length - 1 ] as HTMLElement ).setAttribute( 'tabindex', '0' );
```
bereket-WMDE: @micgro42 I think this line can/ should be Implemented in wikit ?

micgro42: Why do you want to set the focus of the toggle group as a whole instead of one of its buttons (probably the not selected one)?

bereket-WMDE: you're right. It doesn't make sense to put the focus on the whole group. 
I change it to focus on the non-selected button

src/components/QueryBuilder.vue
```diff
@@ -119,6 +119,13 @@ export default Vue.extend( {
 				document
 					.getElementsByClassName( 'querybuilder__condition-wrapper-last' )[ 0 ]
 					.scrollIntoView( { behavior: 'smooth' } );
+
+				const toggle = document
+					.querySelectorAll( '.querybuilder__condition-relation-toggle .wikit-ToggleButton' );
+				const toggleElement = toggle[ toggle.length - 1 ] as HTMLElement;
+
+				toggleElement.setAttribute( 'tabindex', '0' );
```
micgro42: Could you explain why you are giving the button a tabindex?

bereket-WMDE: good catch. this is not needed at all, since the element is already focusable.

micgro42 (approved):

PR #222, opened 2021-03-12
accessibility improvement: trash can should be focussed last

bereket-WMDE:
Bug: [T276942](https://phabricator.wikimedia.org/T276942)

I thought about potential ways to tackle this problem.

1. Typescript manipulation.
      We can add some typescript code to change the behavior of Tab navigation. but this is messy and often leads to bugs in       different browsers. 
2.  We can Explicit add a positive  `Tabindex` values to the `NegationToggle` and `DeleteConditionButton` components. but this is not a good pattern. as can be read [here](https://bitsofco.de/how-and-when-to-use-the-tabindex-attribute/)

> There is almost no reason to ever use a positive value to tabindex, and it is actually considered an anti-pattern. If you’re finding the need to use this value to change the order in which elements become focusable, it is likely that what you actually need to do is change the source order of the HTML elements.

3. change the DOM order and make a few tweaks to the style and then let the natural keyboard navigation do its thing.

**I decided to go with number 3**

### DIFF
--- a/src/components/QueryCondition.vue
+++ b/src/components/QueryCondition.vue
@@ -1,9 +1,5 @@
 <template>
 	<div class="query-condition">
-		<DeleteConditionButton
-			class="query-condition__remove"
-			@click="removeCondition"
-		/>
 		<NegationToggle
 			class="query-condition__toggle-button-group"
 			v-model="negateValue"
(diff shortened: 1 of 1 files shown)


guergana (changes requested):
I like the simple solution but the mobile version doesn't behave the same:

This is how the current deployed version looks like

image

and this is how it looks in this PR. I think the flex needs to be tweaked a little bit more.

image

bereket-WMDE:
@guergana Thank you for the feedback.
I adjusted the styles. can be tested again

guergana (approved):

PR #224, opened 2021-03-12
test: add matrix test to cover most SPARQL generation

micgro42:
This covers almost every permutation of a single condition QueryRepresentation. It doesn't permutate limit and omitLabels.

These tests are expected to be very useful when refactoring the way we create the SPARQL because they can prove that the SPARQL generation of a condition did not change for any given simple refactoring.

It currently generates _18_ ( 2 negation options * 3 property value options * 3 reference options ) test cases per given set of data. It currently includes 3 sets of data: string value, wikibase-item value, and wikibase-item value with subclasses.

**Reviewing the snapshots now is not necessary!**

Those snapshots are only interesting when they _change_.


bereket-WMDE (approved):
@micgro42 Thank you for creating this. 
I really like the idea and its going to be very helpful. Its a very handy way of insuring query integrity

guergana (commented):
Wow. This is pretty cool. I've tested it and when one snapshot doesn't match the feedback is 

image

is there a way of being more explicit and saying that the test failed instead of saying it's obsolete?

micgro42:
Thank you :pray:

PR #226, opened 2021-03-15
refactor: make only locally used method private

micgro42:
This isn't used outside the TripleBuilder class.


guergana (approved):
good catch

PR #228, opened 2021-03-15
refactor: extract method for building the reference patterns

micgro42:
This makes the method much clearer and will make it easier to fix the error about negated conditions that check for references.

Bug: T277308

### DIFF
--- a/src/sparql/PatternBuilder.ts
+++ b/src/sparql/PatternBuilder.ts
@@ -1,7 +1,7 @@
 import TripleBuilder from '@/sparql/TripleBuilder';
 import rdfNamespaces from '@/sparql/rdfNamespaces';
 import { Condition } from '@/sparql/QueryRepresentation';
-import { FilterPattern, Pattern, Triple } from 'sparqljs';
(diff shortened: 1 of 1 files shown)


src/sparql/PatternBuilder.ts
```diff
@@ -15,6 +15,57 @@ export default class PatternBuilder {
...
+		const patterns: Pattern[] = [];
+
+		if ( negate === true ) {
+			patterns.push( {
+				type: 'minus',
+				patterns: [ bgp ],
+			} );
+		} else {
+			if ( condition.referenceRelation === ReferenceRelation.With ||
+				condition.referenceRelation === ReferenceRelation.Without ) {
```
bereket-WMDE: can't we just check `regardless` here ?
`if (condition.referenceRelation !== ReferenceRelation.regardless)`

bereket-WMDE:
apart from a potential minor change, its looks good

src/sparql/PatternBuilder.ts
micgro42: I just moved this around, but you're right, comparing only with regardless is easier to understand.

Done.

bereket-WMDE (approved):

PR #229, opened 2021-03-15
fix: make negated queries handle references correctly 

micgro42:
This makes the method much clearer and will make it easier to fix the error about negated conditions that check for references.

Bug: T277308


micgro42:
Somehow this PR had the title of the previous commit :facepalm:

bereket-WMDE (approved):

PR #230, opened 2021-03-15
refactor: depend less on condition interface

micgro42:
This makes it much more explicit which method depends on what values. It adheres more to the I in SOLID: ISP -> interface segregation principle

### DIFF
--- a/src/sparql/PatternBuilder.ts
+++ b/src/sparql/PatternBuilder.ts
@@ -1,3 +1,4 @@
+import UnitValue from '@/data-model/UnitValue';
 import TripleBuilder from '@/sparql/TripleBuilder';
 import rdfNamespaces from '@/sparql/rdfNamespaces';
 import { Condition } from '@/sparql/QueryRepresentation';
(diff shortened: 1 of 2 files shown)


Ladsgroup:
This looks like a code smell to me (https://pragmaticways.com/31-code-smells-you-must-know/#9_Long_Parameter_List)
> When functions or methods exceed 4+ parameters, then there’s likely 1 of 2 problems going on:
>  there’s more than one thing happening in the function
> the parameters are tightly related and should be grouped together in a data object

Maybe introduce a new value object instead of using condition? Would that make sense?

micgro42:
> This looks like a code smell to me ([pragmaticways.com/31-code-smells-you-must-know/#9_Long_Parameter_List](https://pragmaticways.com/31-code-smells-you-must-know/#9_Long_Parameter_List))
> 
> > When functions or methods exceed 4+ parameters, then there’s likely 1 of 2 problems going on:
> > there’s more than one thing happening in the function
> > the parameters are tightly related and should be grouped together in a data object
> 
> Maybe introduce a new value object instead of using condition? Would that make sense?

I agree that there is a code smell here, but I suspect this code smell is mainly on the side of more than one thing going on and we should split the methods. This will be especially relevant when we start to add the QuantityValuesPatternBuilder as I will do in two PRs. That will introduce even more prefixes. I think it makes more sense to figure out the right refactorings when we have that code, than trying to do that beforehand.

I'm currently trying to get this code into shape that makes these later refactorings easier, so this right here is not intended as its final form, but merely a more malleable one. So having the smells being more obvious is maybe not even a bad thing for now :)

Ladsgroup:
> > This looks like a code smell to me ([pragmaticways.com/31-code-smells-you-must-know/#9_Long_Parameter_List](https://pragmaticways.com/31-code-smells-you-must-know/#9_Long_Parameter_List))
> > > When functions or methods exceed 4+ parameters, then there’s likely 1 of 2 problems going on:
> > > there’s more than one thing happening in the function
> > > the parameters are tightly related and should be grouped together in a data object
> > 
> > 
> > Maybe introduce a new value object instead of using condition? Would that make sense?
> 
> I agree that there is a code smell here, but I suspect this code smell is mainly on the side of more than one thing going on and we should split the methods. This will be especially relevant when we start to add the QuantityValuesPatternBuilder as I will do in two PRs. That will introduce even more prefixes. I think it makes more sense to figure out the right refactorings when we have that code, than trying to do that beforehand.
> 
> I'm currently trying to get this code into shape that makes these later refactorings easier, so this right here is not intended as its final form, but merely a more malleable one. So having the smells being more obvious is maybe not even a bad thing for now :)

I'm sold.

Ladsgroup (approved):

micgro42:
Thank you :pray:  :blush:

PR #232, opened 2021-03-16
refactor: rename private methods to better match what they do

micgro42:
The names were off in a way that made it harder to think about that class.


Ladsgroup (approved):
That's my mess. Sorry.

micgro42:
> That's my mess. Sorry.

Don't worry, naming is hard for everyone! Thank you for approving

PR #234, opened 2021-03-16
Create ParseValueRepository

bereket-WMDE:
Bug: [T277038](https://phabricator.wikimedia.org/T277038)

### DIFF
--- a/src/data-access/FetchParseValueRepository.ts
+++ b/src/data-access/FetchParseValueRepository.ts
@@ -10,19 +10,26 @@ export default class FetchParseValueRepository implements ParseValueRepository {
 		this.endpoint = endpoint;
 	}
 
-	private async parseValue( values: string, datatype: string ):
-	Promise<ParseResult> {
+	public async parseValues( values: string[], datatype: string ):
+	Promise<ParseResult[]> {
 		if ( !values ) {
 			throw new Error( 'The parameter values must not be empty!' );
+		}
+
+		// NOTE: remove this line if/when we decide to supported with multiple values.
+		if ( values.length > 1 ) {
+			throw new Error( 'only one value is supported.' );
 		}
 
 		if ( !datatype ) {
 			throw new Error( 'The parameter datatype must not be empty!' );
 		}
 
+		const concatinateValues = values.join( '|' );
+
 		const params: { [key: string]: string } = {
 			action: 'wbparsevalue',
-			values: values,
+			values: concatinateValues,
 			datatype: datatype,
 			validate: 'true',
 			format: 'json',
@@ -52,19 +59,12 @@ export default class FetchParseValueRepository implements ParseValueRepository {
 			throw new TechnicalProblem( `${data.error.code}: ${data.error.info}` );
 		}
 
-		if ( data.value.precision < 9 ) {
+		const results = data.results;
+
+		if ( results[ 0 ].precision < 9 ) {
 			throw new PrecisionError();
 		}
 
-		const parsedResult: ParseResult = {
-			value: data.value,
-			type: data.type,
-		};
-
-		return parsedResult;
-	}
-
-	public parseValues( values: string, datatype: string ): Promise<ParseResult> {
-		return this.parseValue( values, datatype );
+		return results;
 	}
 }
--- a/src/data-access/FetchParseValueRepository.ts
+++ b/src/data-access/FetchParseValueRepository.ts
@@ -1,0 +1,70 @@
+import ParseValueRepository from '@/data-access/ParseValueRepository';
+import ParseResult from '@/data-access/ParseResult';
+import TechnicalProblem from '@/data-access/errors/TechnicalProblem';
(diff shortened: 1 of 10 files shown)


src/data-access/FetchParseValueRepository.ts
```diff
@@ -0,0 +1,70 @@
...
+
+		if ( data.error ) {
+			throw new TechnicalProblem( `${data.error.code}: ${data.error.info}` );
+		}
+
+		if ( data.value.precision < 9 ) {
+			throw new PrecisionError();
+		}
+
+		const parsedResult: ParseResult = {
```
micgro42: from looking at an [actual result](https://www.wikidata.org/w/api.php?action=wbparsevalue&datatype=time&validate&values=1994-02-08) it would seem that the datastructure that is returned by the API is somewhat different than what is expected by the code here. There seems to be a `results` key involved that contains an array?
But maybe I'm missing something?

src/data-access/FetchParseValueRepository.ts
```diff
@@ -0,0 +1,70 @@
...
+			value: data.value,
+			type: data.type,
+		};
+
+		return parsedResult;
+	}
+
+	public parseValues( values: string, datatype: string ): Promise<ParseResult> {
```
micgro42: if we expect multiple values in our interface/signature (which makes sense to me), then we probably should accept an array of strings and concatenate them ourselves.

src/data-access/FetchParseValueRepository.ts
```diff
@@ -0,0 +1,70 @@
...
+		return this.parseValue( values, datatype );
```
micgro42: What is the idea behind this level of indirection?

tests/unit/data-access/FetchParseValueRepository.spec.ts
```diff
@@ -0,0 +1,119 @@
...
+describe( 'FetchParseValueRepository', () => {
+
+	it( 'parses values with time datatype', async () => {
+		const testEndpoint = 'https://example.com/w/api.php';
+		const repo = new FetchParseValueRepository(
+			testEndpoint,
+		);
+		const value = '1994-02-08';
+		const datatype = 'time';
+		const expectedResult = {
```
micgro42: That looks like the thing that the Repository is supposed to return, but it seems to also be used in line 28 as the mocked API response?

src/data-access/FetchParseValueRepository.ts
bereket-WMDE: fixed

src/data-access/FetchParseValueRepository.ts
bereket-WMDE: makes sense. added support for array of strings.

src/data-access/FetchParseValueRepository.ts
bereket-WMDE: I honestly didn't think too much of that line. I copied most of the structure from `FetchSearchEntityRepository` file. but you're right. it has no purpose here. removed

tests/unit/data-access/FetchParseValueRepository.spec.ts
bereket-WMDE: fixed

src/data-access/FetchParseValueRepository.ts
```diff
@@ -0,0 +1,70 @@
...
+		// NOTE: remove this line if/when we decide to supported with multiple values.
+		if ( values.length > 1 ) {
+			throw new Error( 'only one value is supported.' );
+		}
+
+		if ( !datatype ) {
+			throw new Error( 'The parameter datatype must not be empty!' );
+		}
+
+		const concatinateValues = values.reduce( ( accumulator, currentValue ) => accumulator + '|' + currentValue );
```
micgro42: You can use `.join( '|' )` instead: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/join

src/data-access/ParseResult.ts
```diff
@@ -0,0 +1,11 @@
+export default interface ParseResult {
```
micgro42, line 1: At some point we probably want to rename that as it is really one specific thing and not a generic ParseResult, but let's roll with it for now

src/data-access/FetchParseValueRepository.ts
bereket-WMDE: ah, thanks. thats much cleaner

src/data-access/ParseResult.ts
bereket-WMDE: true true. `ParseValueResult` ? 
its still sounds pretty generic, though

micgro42: Nah, this particular interface is probably `TimeDataValue` and this repo would just return a generic `DataValue` interface. But again, that's something for a different date. Maybe we'll create a tech improvement ticket for this

micgro42 (approved):
\o/

PR #235, opened 2021-03-17
refactor: split base triple regardless of references

micgro42:
This changes how the SPARQL that we generate looks (a bit uglier), but has no functional changes.
But it makes future refactoring and especially reference handling much, _much_ easier.
That will be introduced in the next commit, to keep this commit with the changes to the generated SPARQL as small as possible.


bereket-WMDE (approved):
Nice Job!
Its actually less daunting to understand than it looks :)

micgro42:
Thanks!

PR #236, opened 2021-03-17
create_formatValue_repository

bereket-WMDE:
Bug: [T277057](https://phabricator.wikimedia.org/T277057)

### DIFF
--- a/src/data-access/FetchFormatValueRepository.ts
+++ b/src/data-access/FetchFormatValueRepository.ts
@@ -1,0 +1,56 @@
+import FormatValueRepository from '@/data-access/FormatValueRepository';
+import ParseResult from '@/data-access/ParseResult';
+import TechnicalProblem from '@/data-access/errors/TechnicalProblem';
+
+export default class FetchFormatValueRepository implements FormatValueRepository {
+	private readonly forLanguageCode: string;
+	private readonly endpoint: string;
+
+	public constructor( endpoint: string, languageCode: string ) {
+		this.forLanguageCode = languageCode;
+		this.endpoint = endpoint;
+	}
+
+	private async format( dataValue: ParseResult, propertyId: string ):
+	Promise<string> {
+		const params: { [key: string]: string } = {
+			action: 'wbformatvalue',
+			generate: 'text/plain',
+			datavalue: JSON.stringify( dataValue ),
+			property: propertyId,
+			uselang: this.forLanguageCode,
+			options: JSON.stringify( { lang: this.forLanguageCode } ),
+			format: 'json',
+			formatversion: '2',
+			errorformat: 'plaintext',
+			origin: '*',
+		};
+
+		const url = new URL( this.endpoint );
+		for ( const key in params ) {
+			url.searchParams.set( key, params[ key ] );
+		}
+		let response: Response;
+		try {
+			response = await fetch( url.toString() );
+		} catch ( e ) {
+			throw new TechnicalProblem( 'Network error' );
+		}
+
+		if ( !response.ok ) {
+			throw new TechnicalProblem( `${response.status}: ${response.statusText}` );
+		}
+
+		const data = await response.json();
+
+		if ( data.error ) {
+			throw new TechnicalProblem( `${data.error.code}: ${data.error.info}` );
+		}
+
+		return data;
+	}
+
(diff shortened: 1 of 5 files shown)


src/data-access/FetchFormatValueRepository.ts
```diff
@@ -0,0 +1,56 @@
+import FormatValueRepository from '@/data-access/FormatValueRepository';
+import ParseResult from '@/data-access/ParseResult';
+import TechnicalProblem from '@/data-access/errors/TechnicalProblem';
+
+export default class FetchFormatValueRepository implements FormatValueRepository {
+	private readonly forLanguageCode: string;
+	private readonly endpoint: string;
+
+	public constructor( endpoint: string, languageCode: string ) {
```
micgro42: This signature expects the endpoint as first and the language code as second argument, but [the actual call of the constructor](https://github.com/wmde/query-builder/pull/236/files#diff-bf287fbde737fe72be832168148747edaf7b96dc7b9da182dcf8d9de0b203be4R18-R21) in `src/ServicesFactory` provides them in the opposite order.

tests/unit/data-access/FetchFormatValueRepository.spec.ts
```diff
@@ -0,0 +1,73 @@
...
+	it( 'format date from valueData returned from ParseValueRepository', async () => {
+		const repo = new FetchFormatValueRepository(
+			testEndpoint,
+			testLang,
+		);
+		const expectedResult = 'February 1920';
+
+		window.fetch = jest.fn().mockImplementation( () => Promise.resolve( {
+			ok: true,
+			json: async () => ( expectedResult ),
```
micgro42: What is returned by the API is not `February 1920`, but `{"result":"February 1920"}`

src/data-access/FetchFormatValueRepository.ts
```diff
@@ -0,0 +1,56 @@
...
+
+		if ( data.error ) {
+			throw new TechnicalProblem( `${data.error.code}: ${data.error.info}` );
+		}
+
+		return data;
+	}
+
+	public formatValue( dataValue: ParseResult, propertyId: string ): Promise<string> {
+		return this.format( dataValue, propertyId );
```
micgro42: I'm confused by this indirection again. Could you explain your thoughts behind it?

src/data-access/FetchFormatValueRepository.ts
bereket-WMDE: :( thanks for the catch!

tests/unit/data-access/FetchFormatValueRepository.spec.ts
bereket-WMDE: fixed

src/data-access/FetchFormatValueRepository.ts
bereket-WMDE: yea.. like is said in the previous PR, it was a result of copy pasting. which is not great.
I removed this line from both PRs

tests/unit/data-access/FetchFormatValueRepository.spec.ts
micgro42: Your _expected result_ is not `{"result":"February 1920"}`, but `"February 1920"`. You need to differentiate between what the API returns and what you want your repository to return.

Currently, the repository is actually directly returning the object it receives from the API.

src/data-access/FetchFormatValueRepository.ts
```diff
@@ -0,0 +1,52 @@
...
+			throw new TechnicalProblem( `${response.status}: ${response.statusText}` );
+		}
+
+		const data = await response.json();
+
+		if ( data.error ) {
+			throw new TechnicalProblem( `${data.error.code}: ${data.error.info}` );
+		}
+
+		return data;
```
micgro42: Based on the signature, this is supposed to be a string. But currently, it returns the entire API response, which is a object.

micgro42 (changes requested):

src/data-access/FetchFormatValueRepository.ts
bereket-WMDE: good eye. i guess the test didn't fail here because it was also expecting an object

micgro42 (approved):

PR #238, opened 2021-03-17
refactor: extract separate Pattern builders for string and items

micgro42:
This slices and dices the PatternBuilder and TripleBuilder in a new way. There is now a StringPatternValueBuilder for strings and external IDs and a ItemValuePatternBuilder for wikibase-item.

While these two classes share seemingly some logic, most of that is incidental. The generation of the references filter is however independent from the datatype and shared via the TripleBuilder.

The SyntaxBuilder is basically just a thin abstraction layer on top of sparqljs and depends only on it.

This setup allows us to add a QuantityValuePatternBuilder in the next step and a DateValuePatternBuilder in the future.

**Note that the tests do not change.**

### DIFF
--- a/src/sparql/ItemValueBuilder.ts
+++ b/src/sparql/ItemValueBuilder.ts
@@ -1,0 +1,160 @@
+import PropertyValueRelation from '@/data-model/PropertyValueRelation';
+import { Condition } from '@/sparql/QueryRepresentation';
+import rdfNamespaces from '@/sparql/rdfNamespaces';
+import SyntaxBuilder from '@/sparql/SyntaxBuilder';
+import TripleBuilder from '@/sparql/TripleBuilder';
+import ValuePatternBuilder from '@/sparql/ValuePatternBuilder';
+import { MinusPattern, Pattern, PropertyPath, Term } from 'sparqljs';
+
+export default class ItemValueBuilder implements ValuePatternBuilder {
+	private readonly tripleBuilder: TripleBuilder;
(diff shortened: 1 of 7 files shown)


src/sparql/ItemValueBuilder.ts
```diff
@@ -0,0 +1,158 @@
...
+
+		if ( datatype !== 'wikibase-item' ) {
+			throw new Error( 'Expected datatype wikibase-item, got: ' + datatype );
+		}
+		if ( typeof value !== 'string' ) {
+			throw new Error( 'Unexpected wikibase-item value type: ' + typeof value );
+		}
+		let patterns: Pattern[] = [];
+
+		const statementVariable = SyntaxBuilder.buildVariableTermFromName( 'statement' + conditionIndex );
```
Ladsgroup: Same suggestion from the previous PR. Move SyntaxBuilder to class dependencies to make it clear (and maybe easier to write unit test for later)

Ladsgroup (commented):
Generally looks good. With that suggestion, it's good to go IMO.

src/sparql/ItemValueBuilder.ts
micgro42: Mh, his shouldn't be mocked by unit tests, it is part of this unit on a fundamental level. But making the dependency a bit more clear can't hurt.

micgro42: Done.

Ladsgroup (approved):

PR #240, opened 2021-03-17
test: use LilyPond instead of quantity as limited support example

micgro42:
Since we are now implementing support for the quantity datatype, we should switch the example for limited support to something that is not going to be supported for a long time: LilyPond musical notation syntax.

### DIFF
--- a/cypress/fixtures/wbsearchentities-limited-support.json
+++ b/cypress/fixtures/wbsearchentities-limited-support.json
@@ -1,24 +1,24 @@
 {
-    "searchinfo": {
-      "search": "installed capacity"
-    },
-    "search": [
-      {
-        "id": "P2109",
-        "title": "Property:P2109",
-        "pageid": 23065851,
-        "repository": "wikidata",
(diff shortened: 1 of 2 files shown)


Ladsgroup (approved):
That sounds fringe enough that we won't get around to support it anytime soon.

bereket-WMDE:
Just wondering if the limited support [here](https://github.com/wmde/query-builder/blob/master/cypress/fixtures/wbsearchentities-limited-support.json) should be the same as this one ? I'm not sure how fringe `installed capacity` is

bereket-WMDE (approved):

micgro42:
> Just wondering if the limited support [here](https://github.com/wmde/query-builder/blob/master/cypress/fixtures/wbsearchentities-limited-support.json?rgh-link-date=2021-03-18T10%3A19%3A41Z) should be the same as this one ? I'm not sure how fringe `installed capacity` is

`installed capacity` has `"datatype": "quantity",`, so we are in the process of adding support right now :)

PR #241, opened 2021-03-18
Show netlify URL in a separate step

guergana:


micgro42:
If the URL changes every time the action runs, then a new comment each time would probably be quite spammy. But having it in the logs is really helpful :pray:

bereket-WMDE (commented):
@guergana 
I was just looking into the wikit implementation of this. 
how about doing the same in the QB . 

you can find the implementation in wikit [here](https://github.com/wmde/wikit/blob/master/.github/workflows/comment-deploy-url-to-pr.yaml)

somehow that implementation, doesn't create multiple comments

guergana:
> @guergana
> I was just looking into the wikit implementation of this.
> how about doing the same in the QB .
> 
> you can find the implementation in wikit [here](https://github.com/wmde/wikit/blob/master/.github/workflows/comment-deploy-url-to-pr.yaml)
> 
> somehow that implementation, doesn't create multiple comments

@bereket-WMDE I didn't want to do exactly the same thing they did. I would prefer to take Michael's suggestion because otherwise we also need to make changes in the netlify configuration in case we want to use the hash to generate the URL. If you agree then we can just keep it the logs and forget about the comment for now. :sweat_smile:

Ladsgroup (changes requested):
We shouldn't do this. We are already moving to gerrit.

guergana:
> We shouldn't do this. We are already moving to gerrit.

It's already done. I think there is no damage. At least in the meantime so it's easier for UX people to review.

micgro42 (approved):
Even if we have it just for one day - no harm done. Maybe it can be a template for future Repositories that start at GitHub.

Ladsgroup (approved):
Let's merge it then.

